PR #7, opened 2016-08-17
Add migration API.

mattcollier:


lib/migration.js
```diff
@@ -0,0 +1,484 @@
...
+// steps per progress bar tick call optimization
+var PROCESSED_PER_TICK = 20;
+
+bedrock.events.on('bedrock-cli.init', function() {
+  // setup common migration options
+  bedrock.program
+    .option('--print-objects', 'Print out changed objects.')
+    .option('--progress', 'Print progress.');
+});
+
```
davidlehn: If this tool should require the `--config` option to setup database config and such then something like the following might work:
```js
bedrock.events.on('bedrock-cli.ready', function() {
  if(bedrock.program.config.length === 0) {
    throw new Error('Database setup via the --config option is required.');
  }
});
```
The above doesn't force you to setup the database, but would be a good hint. The config values could be checked here too but I'm not sure how to distinguish between the default values and use specified values.

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+/**
+ * Exported function to run test code via Bedrock.
+ *
+ * @param callback callback when system started. Will not be started if an
+ *   error occurred.
+ */
+api.main = function(callback) {
+  bedrock.events.on('bedrock-cli.ready', function(callback) {
```
mattcollier: @davidlehn added this here.

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+
+var logger = bedrock.loggers.get('app');
+
+var api = {};
+module.exports = api;
+
+// steps per progress bar tick call optimization
+var PROCESSED_PER_TICK = 20;
+
+bedrock.events.on('bedrock-cli.init', function() {
```
davidlehn: This should move into `api.main()` too since otherwise every app using bedrock-mongodb will see these migration options.

mattcollier: @davidlehn done.

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+ *
+ * @param err the error if there was one
+ * @param results the set of results, which are ignored
+ */
+api.complete = function(err) {
+  if(err) {
+    logger.error('MongoDB migration failed:', err);
+    // FIXME improve bedrock shutdown method
+    process.exit(1);
+  } else {
```
dlongley: Else unnecessary (exited above).

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+        }
+        callback(null, database.collections[options.collection]);
+      });
+    },
+    ready: ['collection', function(callback, results) {
+      if('ready' in options) {
+        options.ready({
+          collection: results.collection
+        }, callback);
+      } else {
```
dlongley: Return early then remove unnecessary `else`.

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+                  results.collection.update({_id: record._id}, record,
+                    function(err) {
+                      _processed();
+                      // FIXME: prevent stack overflow
+                      // process.nextTick(function() {
+                      //  callback(err);
+                      // });
+                      callback(err);
+                    });
+                } else {
```
dlongley: Return early and then remove unnecessary `else`.

lib/migration.js
```diff
@@ -0,0 +1,493 @@
...
+      }
+      try {
+        options.update(record, function(err, doUpdate) {
+          if(err) {
+            return callback(err);
+          }
+          // checking for explicit 'false' flag to skip update
+          if(doUpdate !== false) {
+            results.collection.update({_id: record._id}, record, callback);
+          } else {
```
dlongley: Return early and then remove unnecessary `else`.

lib/migration.js
```diff
@@ -0,0 +1,484 @@
...
+  process.exit();
+};
+
+/**
+ * Run function on every element of a table.
+ *
+ * The functions are called in order if present:
+ * 'read': read all objects.
+ * 'update' mutate objects in place. call callback with false to avoid writing
+ * dreturn false can be used to mutate objects in place.
```
dlongley: Is this a typo?

lib/database.js
```diff
@@ -54,6 +54,9 @@ api.localCollection = null;
 // local document ID
 api.localDocumentId = 'local';
 
+// migration API
+api.migration = require('./migration');
```
dlongley: If we think this is an unstable API right now... perhaps we should expose it as `api._migration` until we're more comfortable with it. Thoughts? @mattcollier @davidlehn

dlongley:
One important question to answer that may change what we call the API. After that decision has been made, this LGTM and can be merged.

PR #8, opened 2016-09-02
Check if mongoDB authentication is required.

mattcollier:
The use case for this feature is automated testing where authentication is not enabled on the mongoDB server.  The admin prompt for creating the `bedrock` user is avoided.  There is no need to create mongoDB user accounts under these conditions.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -24,20 +24,23 @@ config.mongodb = {};
 //
 // `mongodb.connectOptions` will override any specified in the URL.
 //
-// format: mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
+// format: mongodb://[username:password@]host1[:port1][,host2[:port2],
+// ...[,hostN[:portN]]][/[database][?options]]
 //
-//config.mongodb.url = 'mongodb://localhost:27017/bedrock_dev';
+// config.mongodb.url = 'mongodb://localhost:27017/bedrock_dev';
 config.mongodb.name = 'bedrock_dev';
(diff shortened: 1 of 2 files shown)


dlongley:
I'm not sure if this is a sufficient check or not. Isn't it possible that you can list databases but you lack the sufficient permission to create them? Are we sure this guess really works the way we want... and do we want to be able to configure whether it's on/off? Are there any other side effects?

If we know we don't need authentication, can't we just set a config var to skip it instead of guessing? We have something similar to this with `adminPrompt`.

mattcollier:
I imagine that if authorization were enabled, and a user was logged in as one user or another, there might be a set of permissions that would give a false positive here, but in this case there has been no attempt made to login at all, and therefore I believe no permissions could be assigned if authorization is enabled.

As for a configuration flag such as `skipAuth` or what have you, if the flag is set incorrectly and `skipAuth` is true when authorization is enabled on the server, the user will find out soon enough when they are presented with authorization errors, and no harm would be done.

I find it remarkable that  `db.admin().serverInfo()` does no provide information about the state of authorization.  One other possibility that I have not pursued is that there is/was a way to determine the command line arguments passed to `mongod`.  I'm not sure if that method would detect if authorization was enabled in the `mongo.conf` file, which happens to be how I have been turning authorization on/off.

dlongley:
Ok. We might want a flag that will perform authentication *anyway* even if it's not strictly required. If we can add that, then I think we'll go ahead and merge this.

mattcollier:
@dlongley flag added.

dlongley:
LGTM, merge.

PR #10, opened 2016-11-04
Replace deprecated `ensureIndex` with `createIndex`.

### LINKED ISSUE #6: Mongo ensureIndex() is deprecated
https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex/

Deprecated since version 3.0.0: db.collection.ensureIndex() is now an alias for db.collection.createIndex().

Using this here:
- https://github.com/digitalbazaar/bedrock-mongodb/blob/master/lib/database.js#L348
- https://github.com/digitalbazaar/bedrock-mongodb/blob/master/lib/database.js#L495

mattcollier:
Fixes #6.

This has been tested against top level issuer and IdP test suites.  No issues detected.

### DIFF
--- a/lib/database.js
+++ b/lib/database.js
@@ -16,7 +16,7 @@
  * is exposed by this module as 'localDocumentId'. The value of 'local' is
  * a JSON object where local properties should be stored.
  *
- * Copyright (c) 2012-2015 Digital Bazaar, Inc. All rights reserved.
+ * Copyright (c) 2012-2016 Digital Bazaar, Inc. All rights reserved.
  */
 var async = require('async');
 var crypto = require('crypto');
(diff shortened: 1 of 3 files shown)


davidlehn:
The `ensureIndex` docs deprecation notice says to use `createIndexes`.  Is `createIndex` ok too?

mattcollier:
@davidlehn Here is the API documentation for ensureIndex: http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html#ensureIndex

Here is createIndex which is identical to ensureIndex: http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html#createIndex

And here is createIndexes which is not the same, the first argument is an array: http://mongodb.github.io/node-mongodb-native/2.1/api/Collection.html#createIndexes

davidlehn:
I was just being literal with the ensureIndex deprecation note, which says "use createIndexes instead".  I imagine createIndex is fine.

PR #14, opened 2017-05-13
Reconnect to MongoDB server on connection failure.

msporny:
Aseta was not reconnecting until I change the reconnect settings to the following. I also enabled TCP keep-alive, to fix a bug that kept cropping up where the TCP connection dies when the database is locked for too long (like during a backup). Reconnect + TCP keep alive seems to have dealt with the problem.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -48,7 +48,12 @@ config.mongodb.connectOptions = {
     journal: true
   },
   server: {
-    auto_reconnect: true
+    socketOptions: {
+      autoReconnect: true, // reconnect on connection failure
+      keepAlive: 15000, // TCP keep-alive every 15 seconds (useful for VPN'd connections)
+    },
+    reconnectInterval: 1000, // attempt reconnect every second
+    reconnectTries: 604800, // attempt to reconnect for a week and then give up
   },
   replSet: {},
   mongos: {}


lib/config.js
```diff
@@ -48,7 +48,12 @@ config.mongodb.connectOptions = {
     journal: true
   },
   server: {
-    auto_reconnect: true
+    socketOptions: {
+      autoReconnect: true, // reconnect on connection failure
```
dlongley: This seems fine. Would be nice to add an autoreconnect test, but it does look like the code will accept either `auto_reconnect` (legacy) or this new fix.

lib/config.js
```diff
@@ -48,7 +48,12 @@ config.mongodb.connectOptions = {
...
+      keepAlive: 15000, // TCP keep-alive every 15 seconds (useful for VPN'd connections)
```
dlongley: I'm not sure the `keepAlive` setting does what the comment says. The documentation says "The number of milliseconds to wait before initiating keepAlive on the TCP socket." Which means maybe we don't want this. The default is 0.

lib/config.js
```diff
@@ -48,7 +48,12 @@ config.mongodb.connectOptions = {
...
+    },
+    reconnectInterval: 1000, // attempt reconnect every second
```
dlongley: The default is already 1000 for this ... so we should probably not specify this here?

lib/config.js
```diff
@@ -48,7 +48,12 @@ config.mongodb.connectOptions = {
...
+    reconnectTries: 604800, // attempt to reconnect for a week and then give up
```
dlongley: I guess this is ok. Hopefully database monitoring tools will pick this up and send alerts since a monitor on the app won't get an alert (maybe) until the app quits a week later.

lib/config.js
dlongley: This `keepAlive` value translates to `keepAliveInitialDelay` in mongodb which ends up getting passed here to node's net socket API:

https://nodejs.org/api/net.html#net_socket_setkeepalive_enable_initialdelay

Here's an explanation of TCP keep alive:

https://stackoverflow.com/questions/1480236/does-a-tcp-socket-connection-have-a-keep-alive

In short, when using keep alive on a TCP connection, it will wait X number of seconds (not set here)  after no activity on the connection, then send an empty ACK (a "keep alive probe"). If there's no reply from the server, it will send another one after `keepAliveInitialDelay` (the value set here) milliseconds. It will keep doing this until the server responds or it runs out of probes (default on linux is 9 probes).

Anyway, I don't think this is doing what we want and we should remove it and figure out if anything else needs to be set to remedy whatever problem we're seeing.

dlongley:
This is quite old -- and doesn't work with the new unified topology which handles connections differently now anyway. Closing.

PR #15, opened 2017-10-05
Update driver

mattcollier:
Summary of some testing done with this branch: https://gist.github.com/mattcollier/b3d7ce61f3756a4dd29ca4517c0f0872


mattcollier:
The only changes directly related to the driver update is moving some of the connect options around and updating some property names.  Without these changes, there were deprecation warnings.

PR #17, opened 2018-06-01
Expose native driver as the `mongo` property.

davidlehn:


mattcollier:
What is the use case for this that cannot be served by `api.client`?  If one is bypassing all the things in this module, why not just require mongodb directly?

davidlehn:
I'm not sure what the right pattern is here.  This is to access the GridFS APIs from another lib.  As a first pass it seemed reasonable to just export mongo just so other libs would certainly use the same native driver as the exposed `client`.

mattcollier:
I see: https://github.com/digitalbazaar/bedrock-document-server/blob/dev/lib/index.js#L165

You've got to get access to the `GridFSBucket` class.

Would it make sense to expose that class instead of `mongo`?  Because ultimately the same client will/should be used right?

davidlehn:
Perhaps?  I just figured a catch-all top level export would leave it open to other (ab)uses.  Mongoose does something similar:
http://mongoosejs.com/docs/api.html#mongoose_Mongoose-mongo

mattcollier:
I think 'Export special driver properties one at a time as needed.'  We are looking at needing a major release of this module to bring us to 3.x drivers.  Hopefully the 3.x driver is better organized.

davidlehn:
I think the 3.x api is just about the same.  I'm trying an alternate approach of just adding a special API for GridFSBuckets: https://github.com/digitalbazaar/bedrock-mongodb/pull/18.  Will remove this PR if that's accepted.

davidlehn:
Using #18 instead.

PR #18, opened 2018-07-05
Add createGridFSBucket API.

davidlehn:
Alternate approach to https://github.com/digitalbazaar/bedrock-mongodb/pull/17.
Add an API to create a GridFSBucket.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -350,17 +350,18 @@ api.createIndexes = (options, callback) => async.each(
     item.fields, item.options, callback), callback);
 
 /**
- * Creates a streaming GridFS bucket interface.
+ * Creates a streaming GridFS bucket instance.
  *
  * By default the writeOptions config value is used for the GridFSBucket
  * writeConcern option.
  *
  * @param options see GridFSBucket documentation
- * @param callback(err, bucket) called once the operation completes.
+ *
+ * @return the new GridFSBucket instance
  */
-api.createGridFSBucket = (options, callback) => {
+api.createGridFSBucket = options => {
   const opts = Object.assign({}, {writeConcern: api.writeOptions}, options);
-  callback(null, new mongo.GridFSBucket(api.client, opts));
+  return new mongo.GridFSBucket(api.client, opts);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -349,6 +349,25 @@ api.createIndexes = (options, callback) => async.each(
...
+ *
+ * By default the writeOptions config value is used for the GridFSBucket
+ * writeConcern option.
+ *
+ * @param options see GridFSBucket documentation
+ * @param callback(err, bucket) called once the operation completes.
+ */
+api.createGridFSBucket = (options, callback) => {
+  let opts = options;
+  if(!('writeConcern' in options)) {
```
mattcollier: could this use `_.defaults` and avoid all the opts/options gyrations?  Although `lodash` is not required in this file, it's already a dependency.

davidlehn: I was micro optimizing for no big reason.  Will simplify.

lib/index.js
```diff
@@ -349,6 +349,20 @@ api.createIndexes = (options, callback) => async.each(
...
+ * Creates a streaming GridFS bucket interface.
+ *
+ * By default the writeOptions config value is used for the GridFSBucket
+ * writeConcern option.
+ *
+ * @param options see GridFSBucket documentation
+ * @param callback(err, bucket) called once the operation completes.
+ */
+api.createGridFSBucket = (options, callback) => {
+  const opts = Object.assign({}, {writeConerrn: api.writeOptions}, options);
```
mattcollier: typo

davidlehn: Oops, fixed.

lib/index.js
```diff
@@ -349,6 +349,20 @@ api.createIndexes = (options, callback) => async.each(
...
+/**
+ * Creates a streaming GridFS bucket interface.
+ *
+ * By default the writeOptions config value is used for the GridFSBucket
+ * writeConcern option.
+ *
+ * @param options see GridFSBucket documentation
+ * @param callback(err, bucket) called once the operation completes.
+ */
+api.createGridFSBucket = (options, callback) => {
```
mattcollier: Why is this async?

davidlehn: Hmm.  I did have a weak reason for this, but it doesn't need to be.  I'll change it.

PR #20, opened 2018-08-29
Callbackify createIndexes API.

mattcollier:
Just wanted to add validation, but callbackified it as well.


davidlehn:
That was a quick merge... and release...
Would be nice to promisify the rest of the API too but not that critical.
Need to update the bedrock peer dep to 1.11.0 due to callbackify usage.

PR #21, opened 2018-08-29
Rollback createIndexes changes.

mattcollier:
It appears that `bedrock.util.callbackify` is not compatible with `async.waterfall`

This throws `callback is not a function` https://github.com/digitalbazaar/bedrock-jobs/blob/master/lib/index.js#L65

I don't want to go down this rabbit hole at this time.


dlongley:
Yeah, the latest version of async.waterfall behaves differently when it gets a promise.

PR #25, opened 2019-07-17
Use sha256 in database.hash API.

mattcollier:
This eliminates sodium-native dep which is a step towards getting bedrock apps to run with node 12.

This switches database has to `sha256`.  It happens that hardware with [SHA Intrinsics](https://gist.github.com/mattcollier/8081d07b8014a53937d8219f6b4a2843) outperforms blake2b as well.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,7 +27,6 @@ const logger = require('./logger');
 const mongo = require('mongodb');
 const mongodbUri = require('mongodb-uri');
 const semver = require('semver');
-const sodium = require('sodium-native');
 const {BedrockError} = bedrock.util;
 
 // load config defaults
@@ -262,9 +261,9 @@ api.hash = key => {
       'Invalid key given to database hash method.',
       'InvalidKey', {key});
   }
-  const hashBuffer = Buffer.allocUnsafe(32);
-  sodium.crypto_generichash(hashBuffer, Buffer.from(key, 'utf8'));
-  return hashBuffer.toString('base64');
+  const md = crypto.createHash('sha256');
+  md.update(key, 'utf8');
+  return md.digest('base64');
 };
 
 /**
--- a/package.json
+++ b/package.json
@@ -27,8 +27,7 @@
     "mongodb-uri": "^0.9.7",
     "progress": "~1.1.8",
     "prompt": "~0.2.14",
-    "semver": "^5.5.0",
-    "sodium-native": "^2.2.3"
+    "semver": "^5.5.0"
   },
   "peerDependencies": {
     "bedrock": "^1.11.0"


mattcollier:
This of course is a breaking change and would be a major 6.x release.

dlongley:
Yuck. The sooner we don't have to do this at all the better. Approving...

davidlehn:
Some thoughts:
- I understand that sodium-native is an issue with Node 12 today, but I assume that will be fixed at some point.  Do we need 12 for something else that makes this urgent?
- What's the real world difference due to those hashing speeds?  Does that have a big effect at the app level?
- I'm a bit concerned that we have to have certain hardware with special features to make this sort of thing performant.  It looks like all the CPUs I use daily would take a performance hit with this hash change.  (Though perhaps it doesn't matter and is just an overall app minor loss?)  It's likely other users of this also may not be able to choose particular hardware.
- Ideally we'd just not need to worry about this and mongo features would deal with it.  Though I suppose we need to benchmark if that is fast enough vs this custom code.

mattcollier:
> Some thoughts:
> 
> * I understand that sodium-native is an issue with Node 12 today, but I assume that will be fixed at some point.  Do we need 12 for something else that makes this urgent?

We want the smallest possible docker container which means using the alpine base image which means native dependencies are problematic.

> * What's the real world difference due to those hashing speeds?  Does that have a big effect at the app level?

The difference between one hash and another at this level would be lost in the standard deviation.
Node sha256 @ 9400 ops/sec (no intrinsics)
Node sha256 @ 29800 ops/sec ([with intrinsics](https://github.com/digitalbazaar/crypto-bench#nodejs-10140-amd-epyc-processor--3ghz)) 
sodium-blake2(generichash) @ 17900 ops/sec.
https://github.com/digitalbazaar/crypto-bench#nodejs-1130-intelr-coretm-i7-3610qm-cpu--230ghz

> * I'm a bit concerned that we have to have certain hardware with special features to make this sort of thing performant.  It looks like all the CPUs I use daily would take a performance hit with this hash change.  (Though perhaps it doesn't matter and is just an overall app minor loss?)  It's likely other users of this also may not be able to choose particular hardware.

We're not making this change to try to achieve benefit from intrinsics.  We are making this decision to eliminate native dependencies.  Also, we'll be better off if we ever have to deal with FIPS compliance.

> * Ideally we'd just not need to worry about this and mongo features would deal with it.  Though I suppose we need to benchmark if that is fast enough vs this custom code.

Agreed, eliminating this all together is the proper response.  As of earlier today we know that we will need to upgrade to the mongo 3.x driver to eliminate the index key size restriction with the current driver.      https://github.com/digitalbazaar/bedrock-mongodb/issues/26

davidlehn:
> > * What's the real world difference due to those hashing speeds?  Does that have a big effect at the app level?
> 
> The difference between one hash and another at this level would be lost in the standard deviation.
> Node sha256 @ 9400 ops/sec (no intrinsics)
> Node sha256 @ 29800 ops/sec ([with intrinsics](https://github.com/digitalbazaar/crypto-bench#nodejs-10140-amd-epyc-processor--3ghz))
> sodium-blake2(generichash) @ 17900 ops/sec.
> https://github.com/digitalbazaar/crypto-bench#nodejs-1130-intelr-coretm-i7-3610qm-cpu--230ghz
> 

I was trying to ask how big the hashing time is on a profile of an actual app vs just the raw benchmarks.  How noticeable will a ~50% or ~150% hashing cost change be?

mattcollier:
@davidlehn wrt the app level effect of this change I said: "The difference between one hash and another at this level would be lost in the standard deviation."

My read on this is that if we had some app level operation involving one hashing operation that we could do at 1 op/sec.  Changing the hashing algo from blake2 to sh256 would result in that operation taking ~1/10000 (one ten thousandth) of a second longer.

This entire operation is already redundant in the common case.  That is where the data in the index fields is less than the 1K limit.

This change allows us to move forward more quickly and easily with our containerization strategy.

As soon as we can upgrade to the mongo 3.x driver, this operation will become redundant in every case.

aljones15:
This looks like it is good to go, but not sure I should be the one Rebase and merging this.
@mattcollier @davidlehn

mattcollier:
@aljones15 thanks for the reminder.  This will be a major release and is going to take some time and effort to get integrated into things.  And it needs to be done carefully so as not to break any existing systems.  Which is to say, there's no hurry on this.

wrt node crypto vs forge.  All the crypto in forge is pure js and in most cases is significantly slower than the native implementations in `crypto` (OpenSSL).  See the performance difference between crypto and forge for sha256 here: https://github.com/digitalbazaar/crypto-bench#nodejs-1130-intelr-coretm-i7-3610qm-cpu--230ghz

dlongley:
@aljones15,

> Although I suppose forge uses lib-sodium.

Just a quick note -- forge doesn't use lib-sodium; it has no dependencies. But we definitely want to use node's crypto implementation here; it's much faster and this module only runs on node (not in the browser).

PR #29, opened 2019-10-30
Update to latest bedrock events API.

davidlehn:


lib/index.js
```diff
@@ -74,17 +75,16 @@ const idGenerators = {};
 // load test config
 bedrock.events.on('bedrock.test.configure', () => require('./test.config'));
 
-bedrock.events.on('bedrock.init', init);
+bedrock.events.on('bedrock.init', promisify(init));
 
-bedrock.events.on('bedrock.start', callback => bedrock.events.emit(
-  'bedrock-mongodb.ready', callback));
+bedrock.events.on('bedrock.start',
```
mattcollier: Style issue here, i guess both params move down.

Issue #19, opened 2018-08-20
Update to 3.x driver


mattcollier:
MongoDB 4.x requires the nodejs v3.x driver: https://docs.mongodb.com/manual/release-notes/4.0/#drivers

The 3.x driver is also required to access some of the latest mongodb 3.x additions.

Here's the list of breaking changes in the 3.x driver: https://github.com/mongodb/node-mongodb-native/blob/3.0/CHANGES_3.0.0.md

One of the big ones is that the `find` API has moved the `projection` parameter into `options.projection`.

Our code base will also be impacted by the changes in the bulkWrite APIs.  I would categorize these changes as improvements because the error handling in the 2.x bulkWrite APIs is really wonky which required silly tactics like this: https://github.com/digitalbazaar/bedrock-ledger-storage-mongodb/blob/master/lib/LedgerEventStorage.js#L134

dmitrizagidulin:
Addressed by PR #33, closing.

PR #33, opened 2020-05-20
Driver3

aljones15:
As of _June 3 2020_  important top level app(s) are running just fine with this.

This PR is breaking for the following libraries:

- [x] https://github.com/digitalbazaar/bedrock-kms/pull/23
- [x] https://github.com/digitalbazaar/bedrock-edv-storage/pull/43
- [x] https://github.com/digitalbazaar/bedrock-session-mongodb/pull/8
- [x] https://github.com/digitalbazaar/bedrock-identity/pull/34

There are multiple other Draft PRs awaiting this release:

- [x] https://github.com/digitalbazaar/bedrock-vc-issuer/pull/37
- [x] https://github.com/digitalbazaar/bedrock-ssm-mongodb/pull/6
- [x] https://github.com/digitalbazaar/bedrock-zcap-storage/pull/3
- [x] https://github.com/digitalbazaar/bedrock-authn-token/pull/17
- [x] https://github.com/digitalbazaar/bedrock-profile/pull/12
- [x] https://github.com/digitalbazaar/bedrock-account/pull/14
- [x] https://github.com/digitalbazaar/bedrock-permission/pull/22
(check mark means CI is passing)

any library that passes 3 parameters to mongo's `find` function (ex: `find(query, options, fields)`) will be broken after this release.

Ensure all of these deprecated methods are not being used:
- [x] Db.prototype.authenticate
- [x] Db.prototype.logout
- [x] Db.prototype.open
- [x] Db.prototype.db
- [x] Db.prototype.close
- [x] Admin.prototype.authenticate
- [x] Admin.prototype.logout
- [x] Admin.prototype.profilingLevel
- [x] Admin.prototype.setProfilingLevel
- [x] Admin.prototype.profilingInfo
- [x] Cursor.prototype.nextObject
- [ ] note: all other draft PRs must be retested with these new options

Issues Addressed:
https://github.com/digitalbazaar/bedrock-mongodb/issues/19 Update to Driver 3.x
https://github.com/digitalbazaar/bedrock-mongodb/issues/35 `useUnifiedTopology: true`
https://github.com/digitalbazaar/bedrock-mongodb/issues/36 runOnceAsync removed
https://github.com/digitalbazaar/bedrock-mongodb/issues/34 remove idGen & migration


lib/index.js
```diff
@@ -535,7 +548,7 @@ api.isDuplicateError = err => (api.isDatabaseError(err) &&
  *
  * @return true if the error is a duplicate key error, false if not.
  */
-api.isDatabaseError = err => (err && err.name === MDBE_ERROR);
+api.isDatabaseError = err => (err && MDBE_ERRORS.includes(err.name));
```
dlongley, line 551: If this is changing -- we'll need to change the docs (see above that it refers only to a duplicate key error) and check every place it is used.

lib/idGenerator.js
```diff
@@ -247,7 +247,7 @@ function _loadIds(idGenerator, callback) {
         query[localKey] = id.local;
         const update = {$set: {}};
         update.$set[localKey] = localId.toHex();
-        return database.localCollection.update(
+        return database.localCollection.updateMany(
```
dlongley, line 250: At a cursory glance, this seems like it should `updateOne` ... why is it `updateMany`?

aljones15: looks like I was not sure if the query updated one or more records here. I think you are correct here though just was not sure if `query[globalKey]` was a query for one or more. I'm assuming I'm was wrong on this one?

lib/index.js
aljones15: Just so you know the original error code `MDBE_ERROR` is not exclusive to a DuplicateKey error not to mention that `isDatabaseError` has all types of weird naming conflicts if intended solely to catch a duplicate key error so the docs were probably wrong to start with. Changed it here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/3a25e58b4407b17872aa270c83a7e712cd28b2f8

mattcollier:
Deprecation warning when using this new stuff:
```
(node:25866) DeprecationWarning: current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.
```

mattcollier:
Need to replace use of `runOnceAsync` with `runOnce`.

Thanks to new node 14 feature `--trace-warnings` figured out where this one was coming from!
```
(node:26039) DeprecationWarning: runOnceAsync() is deprecated. Use runOnce() instead.
    at deprecatedCallbackified (util.js:210:5)
    at initDatabase (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/lib/index.js:146:64)
    at runTask (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1634:13)
    at /home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1572:13
    at processQueue (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1582:13)
    at Object.auto (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1568:5)
    at init (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/lib/index.js:145:9)
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at init (internal/util.js:296:12)

```

aljones15:
> Need to replace use of `runOnceAsync` with `runOnce`.
> 
> Thanks to new node 14 feature `--trace-warnings` figured out where this one was coming from!
> 
> ```
> (node:26039) DeprecationWarning: runOnceAsync() is deprecated. Use runOnce() instead.
>     at deprecatedCallbackified (util.js:210:5)
>     at initDatabase (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/lib/index.js:146:64)
>     at runTask (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1634:13)
>     at /home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1572:13
>     at processQueue (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1582:13)
>     at Object.auto (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/node_modules/async/dist/async.js:1568:5)
>     at init (/home/matt/bedrock-dev/bedrock-rsvp/test/node_modules/bedrock-mongodb/lib/index.js:145:9)
>     at internal/util.js:297:30
>     at new Promise (<anonymous>)
>     at init (internal/util.js:296:12)
> ```

Addressed the runOnceAsync issue here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/24bb26da128b8be62f3178bf9f322e29bb4b0823

the code in that commit does not address removing `db.authneticate` yet.

aljones15:
> Deprecation warning when using this new stuff:
> 
> ```
> (node:25866) DeprecationWarning: current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.
> ```

(node:16700) DeprecationWarning: The option `autoReconnect` is incompatible with the unified topology, please read more by visiting http://bit.ly/2D8WfT6

If I can get around to this I will, but the deprecation warning is not fatal.


Docs on the newer topology method here:
http://mongodb.github.io/node-mongodb-native/3.3/reference/unified-topology/

lib/index.js
```diff
@@ -686,14 +701,27 @@ function _openDatabase(options, callback) {
       const username = config.username;
       const password = config.password;
       let opts = config.authentication;
+      const {MongoClient, Server} = mongo;
       if(_usesRoles(results.serverInfo)) {
         // authenticate against shared db
-        opts = bedrock.util.extend({}, opts, {authdb: config.name});
-        return db.authenticate(username, password, opts, callback);
+        opts = bedrock.util.extend({}, opts, {authSource: config.name});
```
dlongley, line 707: This would do the same, right? We should phase out `bedrock.util.extend` where new language options are simpler.

```suggestion
        opts = {...opts, authSource: config.name};
```

dlongley:
This link has more information (more than the 3.3 reference) on the unified topology changes: https://mongodb.github.io/node-mongodb-native/3.4/reference/unified-topology/

dlongley:
So regarding `autoReconnect`, it looks like previous behavior would have used these settings:

```
   * - **autoReconnect**
     - Server
     - boolean
     - true
     - Enables autoReconnecting when using a standalone server in the legacy topology setup
   * - **reconnectTries**
     - Server
     - integer
     - 30
     - When ``autoReconnect`` is enabled, the server attempt to reconnect #times.
   * - **reconnectInterval**
     - Server
     - integer
     - 1000
     - When ``autoReconnect`` is enabled, the server will wait # milliseconds between retries.
```

https://github.com/mongodb/node-mongodb-native/blob/71eee740721b6bfc8e0f7656e1d2dc418eeea7da/docs/guide/connection-settings.txt#L226-L240

So, with `autoReconnect: true`, the code would effectively implement a loop whereby an attempt to connect to the server would be performed every second for up to 30 seconds. With `useUnifiedTopology`, instead, the docs say a loop like this:

```js
function executeOperation(topology, operation, callback) {
  const readPreference = resolveReadPreference(operation);
  topology.selectServer(readPreference, (err, server) => {
    if (err) {
      // This error is most likely a "Server selection timed out after Xms"
      return callback(err);
    }

    // checks a connection out of the server to execute the operation, then checks it back in
    server.withConnection(conn => operation.execute(conn, callback));
  })
}
```

Will be executed for up to `serverSelectionTimeoutMS` which defaults to 30 seconds. It's unclear how often an attempt will be made to get a connection (not clear if this is modulated by 1 second intervals or something reasonably equivalent to the other method), but it doesn't really matter to us.

Given all this, I think `{useUnifiedTopology: true, serverSelectionTimeoutMS: 30000, autoReconnect: false}` is roughly equivalent to `{autoReconnect: true}`. So let's turn off `autoReconnect` and, ideally, just remove the option from our defaults moving forward.

lib/index.js
aljones15: This was addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/1c76605ece2b2c3a295ccfe074866b716281e0de

davidlehn:
If this package, or our packages in general, are using newer database features, please update the config.js required server version:
```js
config.mongodb.requirements.serverVersion = '>=3.0';
```

lib/index.js
```diff
@@ -131,7 +144,7 @@ function init(callback) {
   }
 
   async.auto({
-    initDatabase: callback => callbackify(bedrock.runOnceAsync)(
+    initDatabase: callback => callbackify(bedrock.runOnce)(
```
davidlehn, line 147: This change means the bedrock peer dependency needs to be bumped to at least ^3.0.0.

aljones15: I changed it to `3.x` here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/4a8f8e5f397cf3594a3c9417b7dfbfed69baae2d

is that a legit peerDep semantic version option? I know that `^3.0.0` means any minor release of 3, but I saw 3.x in `bedrock-vc-issuer` so I used it.

lib/index.js
```diff
@@ -723,16 +746,38 @@ function _connect(options, callback) {
...
+ *
+ * @param {object} options - Options to use.
+ * @param {object} options.auth - user and password credentials.
+ * @param {object} options.opts - Options for the MongoClient.
+ * @param {Function} options.callback - A callback function.
+ * @param {Server} options.server - The mongo server to connect to.
+ *
+ * @returns {Promise} The result of the connect.
+*/
+function _loginUser({auth, opts, callback, server}) {
```
dlongley, line 769: Since this returns a promise, we should mark it `async`, it helps with stack traces.

```suggestion
async function _loginUser({auth, opts, callback, server}) {
```

aljones15: I really can't believe I missed that one, but I made the correct here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/4733ee5f8dc3799abfe26b0bb721325cdea81a1d I guess having the callback confused me?

test/test.config.js
```diff
@@ -7,8 +7,14 @@ const {config} = require('bedrock');
 const path = require('path');
 
 // MongoDB
-config.mongodb.name = 'bedrock_authn_token_test';
+config.mongodb.name = 'bedrock_mongodb_test';
+config.mongodb.username = 'root';
+config.mongodb.password = 'root';
 config.mongodb.dropCollections.onInit = true;
 config.mongodb.dropCollections.collections = [];
+config.mongodb.authentcation = {
```
dlongley, line 15:
```suggestion
config.mongodb.authentication = {
```

package.json
```diff
@@ -26,14 +26,14 @@
...
-    "mongodb": "~2.2.31",
+    "mongodb": "^3.5.7",
     "mongodb-uri": "^0.9.7",
     "progress": "~1.1.8",
     "prompt": "~0.2.14",
     "semver": "^6.3.0"
   },
   "peerDependencies": {
-    "bedrock": "1.11.0 - 3.x"
+    "bedrock": "3.x"
```
dlongley, line 36: I think this is equivalent, but for consistency with our other packages:

```suggestion
    "bedrock": "^3.0.0"
```

test/test.config.js
dlongley: Since this had a typo ... can we be sure the tests were working?

package.json
davidlehn: Might as well make it ^3.1.1 while we're at it.

lib/index.js
davidlehn: They expand the same:
https://www.npmjs.com/package/semver#x-ranges-12x-1x-12-https://www.npmjs.com/package/semver#caret-ranges-123-025-004

package.json
aljones15: @dlongley `3.x` is consistent with our other packages: https://github.com/digitalbazaar/bedrock-vc-issuer/blob/5b291c5b101f5bf892931387ab91a2f9d728d08c/package.json#L47-L63 but I will change it to `^3.1.1` shortly.

aljones15:
> Approving ... left a few nits. My main concern is over whether the authentication test is working since there was a typo in the test config. If we can confirm that's ok, I think we're good here.

your commit for the typo did break the auth tests: (node:3668) UnhandledPromiseRejectionWarning: DatabaseError: Could not initialize database.

which is strange as I added a logger and saw the tests hit the login and pass with: `config.mongodb.forceAuthentication = true;` set to true. It is probably the case that the `authMechanism` is the problem.

package.json
aljones15: Version addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/33/commits/4f74e384eac4ee82c5597f7cfe87ebcd982cabeb

dlongley:
I'm happy for this to move ahead when others are as well.

.github/workflows/main.yml
```diff
@@ -5,20 +5,20 @@ on: [push]
...
     runs-on: ubuntu-latest
-    services:
-      mongodb:
-        image: mongo:3.6
-        ports:
-          - 27017:27017
     strategy:
       matrix:
         node-version: [10.x, 12.x, 14.x]
+        mongodb-version: ['3.6', '4.0', '4.2']
```
mattcollier, line 11: @davidlehn I believe you added this.  As I mentioned before, there is no expectation that this is backwardly compatible with 3.x, and I don't think the index field length thing landed until v4.2.  I wrote all that up somewhere.  At any rate, it is my assertion that we only need to test here with 4.2.

davidlehn: Yes, I added the matrix since it was only testing with 3.6 before: https://github.com/digitalbazaar/bedrock-mongodb/pull/38.

I had been thinking that this module might still work with 3.6.  But other modules might be using the newer features.  Is the suggestion that we just require 4.2+ everywhere and not even test anything else?

If so, my suggestion would be to leave the matrix style there with just 4.2 for now.  4.4 can be added easily later.  (Don't change back to the single mongo service style.)  And just to make sure... update lib/config.js serverVersion to ">= 4.2" so the system will not even start with an older server.

mattcollier: +1 to keeping matrix style in and seting the serverVersion to 4.2+

dmitrizagidulin: +1, @mattcollier's comment about 4.2-only is addressed by PR #40, marking this thread as resolved.

PR #37, opened 2020-06-01
Add default writeOption for forceServerObjectId.

mattcollier:
Docs on this option here: https://mongodb.github.io/node-mongodb-native/3.6/api/Collection.html#insertOne

More description: https://jira.mongodb.org/browse/NODE-1160

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -54,7 +54,8 @@ config.mongodb.connectOptions = {
 config.mongodb.writeOptions = {
   j: true,
   w: 'majority',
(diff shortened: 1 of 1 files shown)


davidlehn:
I think this is fine?  Is this preventative or was there a reason it needs to be this way?

aljones15:
I went ahead and rebased this in as it does not seem to be breaking and setting the `_id` on server does not matter I believe because we generate ids client side and will be using them to find the document.

PR #38, opened 2020-06-02
Add test version matrix for mongodb.

davidlehn:
Previously just testing this on mongodb 3.6.  I put in 3.6, 4.0, and 4.2 here.  Is testing on 4.0 needed?


aljones15:
Not sure, but thanks for adding this. This answers the question of what happens in `mongo 4.2`

PR #40, opened 2020-06-04
Update required server version to >=4.2.

davidlehn:
- Only test with 4.2.
- Keeping mongo version matrix to more easily add new versions.

Based on comment: https://github.com/digitalbazaar/bedrock-mongodb/pull/33#discussion_r434886994

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -5,4 +5,4 @@ const bedrock = require('bedrock');
 require('bedrock-mongodb');
 
 require('bedrock-test');
-bedrock.start();
+bedrock.start().catch(err => console.error(err));
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         node-version: [10.x, 12.x, 14.x]
-        mongodb-version: ['3.6', '4.0', '4.2']
+        mongodb-version: ['4.2']
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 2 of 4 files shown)


aljones15:
@davidlehn just so you know have `serverVersion` set to `4.2` caused the entire test project to crash on my computer with could not initialize database. Lowering it to 4.0 solved the issue for me. So I have no problem with the tests being done on `4.2` on github actions, but it might be better to stick with `>=4.0` as neither my mac or my linux box has mongo `4.2` yet. Oh yeah and good work! COOKIES!!!

davidlehn:
@aljones15 Thank you for confirming the version check code works.  The error output was poor. 
 There is a error cause chain saying the server version check failed.  I add a quick patch to help show that.

The purpose of this patch is to default to 4.2 since, as @mattcollier said, our other projects will start to use 4.2 features.  Obviously installing 4.2 will be required if 4.2 features are used.  This module itself doesn't seem to require 4,2 though.  The version bump is just a default to avoid having to specify the required version in many other apps.

Others, please chime in.  What minimum version should this module specify?

mattcollier:
@aljones15 +1 to what @davidlehn said, we are going to start using 4.2 only features in our stuff.  So that's going to be a requirement.

Here's some instructions on running mongodb in docker if it helps.  I know you've also got datasets to be concerned about.  So hopefully the upgrade is not too painful for you.

test/test.js
```diff
@@ -5,4 +5,4 @@ const bedrock = require('bedrock');
 require('bedrock-mongodb');
 
 require('bedrock-test');
-bedrock.start();
+bedrock.start().catch(err => console.error(err));
```
dmitrizagidulin, line 8:
```suggestion
bedrock.start().catch(console.error);
```

PR #41, opened 2020-06-09
Do not set db if no client.

aljones15:
Already found a bug. Because we are using callbacks if the connect fails then client is undefined. If client is undefined then `client.db` is also undefined. This results in getting an object undefined error instead of the connection error you would expect. So we check if `client` exists before calling on `client.db` then if there is no client we pass the connection error to the callback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,6 @@ connectOptions.tls = true;
 // the `authSource` option replaces the older `authDB` option
 // it should be specified or else it will be the `mongodb.name`
 connectOptions.authSource = 'my_provider_auth_db'; 
-connectOptions.authSource = 'admin';
 ```
 
 ## Requirements
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ const {config} = require('bedrock');
 config.mongodb.username = 'me';
 config.mongodb.password = 'password';
 const {connectOptions} = config.mongodb;
-// this optional and only required if connecting to a replicaSet
+// optional, only required if connecting to a replicaSet
 connectOptions.replicaSet = 'my_provider_replica_set';
 // optional, but required in production by many providers
 connectOptions.ssl = true;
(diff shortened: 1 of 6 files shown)


davidlehn:
Is that useNewUrlParser change just a patch change too?  Does it need a changelog entry?

README.md
```diff
@@ -60,29 +60,35 @@ bedrock.events.on('bedrock-mongodb.ready', function(callback) {
...
+MongoDB's documentation offers tons of great examples on how to authenticate
+using a myriad number of connection strings.
 
-## Configuration
+[Mongo Node 3.5 Driver connect docs](http://mongodb.github.io/node-mongodb-native/3.5/tutorials/connect/)
 
-For documentation on database configuration, see [config.js](./lib/config.js).
+[Mongo Node 3.5 Driver atlas docs](https://docs.mongodb.com/drivers/node#connect-to-mongodb-atlas)
+
+You can also connect to access-enabled mongo servers using some small changes to the
```
msporny, line 75:
```suggestion
You can also connect to mongo servers with authorization enabled using some small changes to the
```

README.md
```diff
@@ -60,29 +60,35 @@ bedrock.events.on('bedrock-mongodb.ready', function(callback) {
...
+const {config} = require('bedrock');
+config.mongodb.username = 'me';
+config.mongodb.password = 'password';
+const {connectOptions} = config.mongodb;
+// optional, only required if connecting to a replicaSet
+connectOptions.replicaSet = 'my_provider_replica_set';
+// optional, but required in production by many providers
+connectOptions.ssl = true;
+// optional, only required if your provider requires tls 
+connectOptions.tls = true;
```
msporny, line 87: I don't understand the difference between the ssl and tls options? I would think they're the same thing? Would be helpful if the difference was pointed out in the comments.

README.md
```diff
@@ -60,29 +60,35 @@ bedrock.events.on('bedrock-mongodb.ready', function(callback) {
...
+// the `authSource` option replaces the older `authDB` option
+// it should be specified or else it will be the `mongodb.name`
```
msporny, line 89:
```suggestion
// `authSource` specifies the database containing the authorization information
// if not specified, `authSource` will be set to the value of `mongodb.name`
// the `authSource` option replaces the older `authDB` option
```

PR #49, opened 2020-07-04
Use projection test & add test to ensure projection is working.

aljones15:
Sorry about the oversite, I noticed that I also got the second parameter to `findOne` incorrect even in the test suite for for this project so I corrected it and added a test which should ensure project is working in the future. I am leaving this as a draft until I can ensure that mongo apis are accepting the correct parameters.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -55,7 +55,10 @@ config.mongodb.connectOptions = {
 config.mongodb.writeOptions = {
   j: true,
   w: 'majority',
+  // as mongo 4.2 this is no longer used by updateOne or updateMany
+  // FIXME remove this once all dependencies have been upgraded
   multi: true,
+  // this is used by insert methods
   forceServerObjectId: true,
 };
 /* DEPRECATED */
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
@@ -52,7 +52,7 @@ jobs:
           - 27017:27017
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
--- a/package.json
+++ b/package.json
@@ -39,8 +39,8 @@
     "lib": "./lib"
   },
   "devDependencies": {
-    "eslint": "^6.6.0",
-    "eslint-config-digitalbazaar": "^2.0.1"
+    "eslint": "^7.4.0",
+    "eslint-config-digitalbazaar": "^2.5.0"
   },
   "engines": {
     "node": ">=8"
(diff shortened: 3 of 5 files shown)


lib/config.js
```diff
@@ -55,7 +55,7 @@ config.mongodb.connectOptions = {
 config.mongodb.writeOptions = {
   j: true,
   w: 'majority',
-  multi: true,
+  // this is used by insert methods
```
mattcollier, line 58: @aljones15 can you describe why multi is being removed?  I have no idea what this is about.

aljones15: The issue here is that the newer mongo api uses `updateOne` and `updateMany`. The later handles what the multi-option used to, While multi is still a valid option it is not longer in the `mongo 3.5` api for these methods: 

http://mongodb.github.io/node-mongodb-native/3.5/api/BulkOperationBase.html

http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#insertOne

http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#insertMany

http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#updateOne

http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#updateMany

basically this seems to be a left over from pre-mongo 2.0. `multi` is valid in `Collection.update`: http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#update

however that is deprecated. 

it is not even used in `Collection.insert` either: http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#insert

which is also deprecated.

BulkWrite does not seem to use it either. So basically a single deprecated `update` method is the only place this option is used. As we are removing deprecated method such as `update` it is probably better to remove `multi` too`. It also appears to `BulkOperations` automatically operate on multi and do not specify `multi` as an option in the api. I recommend looking at the api docs from a high-level and searching for `multi` in the Collections and BulkWrite apis.

mattcollier: Thanks for the information.  Based on the documentation for the v3.5 update API, multi *is* an important option.  Which means: `Update one/all documents with operation.`  Which is unchanged from the [2.2 API](http://mongodb.github.io/node-mongodb-native/2.2/api/Collection.html#update)  I take this to mean that when `multi: false` (the default), it functions as an `updateOne` and when `multi: true` it's an `updateMany`.

I do not know how many instances we have where `update` is used in conjunction with `database.writeOptions`.  Has replacing `update` with `updateOne` or `updateMany` to avoid deprecation notices been a part of the MongoDB upgrade process?

Whatever the case, removing `multi` from `writeOptions` now constitutes a breaking change.  This is because it is still legitimate to have an `update` call that uses `writeOptions` and effectively changing `multi` to false by removing it here may break existing code that relies on `multi: true`.

I do not think anyone is going to be interested in making another massive MongoDB upgrade at this time.  I recommend we do not remove multi, but instead make an issue and a TODO for the deprecated feature.  Hopefully we can keep track of the deprecation issue so that when the next major release comes we can incorporate those changes.

aljones15: I removed the commit to removed `multi` and reverted to the commit with the FIXME: https://github.com/digitalbazaar/bedrock-mongodb/pull/49/commits/3fa2e30d854eb7cb9c1045779928da6bec44b227

mattcollier (approved):

PR #52, opened 2020-08-24
Remove local collection logic.

aljones15:
The test project for this is passing and this has been tested with 2 other projects with no issues.

This removes the local database, localClient, and updates _createUser to work with the default mongodb config.

new create Admin Prompt:

Screen Shot 2020-08-26 at 3 06 40 PM

how to test the Admin Prompt:

1. [enable access management for](https://docs.mongodb.com/manual/tutorial/enable-authentication/) your mongodb server.
2. ensure you have an admin added to the server.
3. enter a new username and password  into the `test.config.js` file in `/test`
4. add connectOptions.authSource into the `test.config.js` file in `/test` (this will probably be `admin`)
5. run the test project
6. You should see the Admin Prompt in the picture above
7. After entering the admin username, password, and optional authentication database the test suite should pass
8. the next run of the test suite should pass with out an admin prompt
9. congrads you have added a new user to your authSource

Note: please only make changes to `test/test.config.js` if necessary.
Note: If you want to you can use environment variables for the username, password, and authSource. You should see the code for that in `test/test.config.js`

https://github.com/digitalbazaar/bedrock-mongodb/blob/07ce9983b71670fafa87424ace8b66a9082f9786/test/test.config.js#L14-L23

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -47,6 +47,8 @@ config.mongodb.connectOptions = {
   serverSelectionTimeoutMS: 30000,
   autoReconnect: false,
   useNewUrlParser: true,
+  // the db to authenticate against
+  authSource: undefined
 };
 config.mongodb.writeOptions = {
   j: true,
--- a/lib/config.js
+++ b/lib/config.js
@@ -67,5 +67,5 @@ config.mongodb.requirements = {};
 // server version requirement with server-style string
 config.mongodb.requirements.serverVersion = '>=4.2';
 
-// this is used by _createUser to add a user as an admin
-//config.mongodb.collection = 'admin-collection';
+// this is used by _createUser to add a user as an admin to a collection
+// config.mongodb.collection = 'admin-collection';
--- a/lib/config.js
+++ b/lib/config.js
@@ -14,10 +14,6 @@ config.mongodb = {};
 // credentials, and any server options will be parsed from the URL and override
 // the broken-down configuration options set here.
 //
-// If `config.mongodb.local.url` is not set, the connection URL for the local
-// database will be assembled from `config.mongodb.local.name`,
-// `config.mongodb.host` and `config.mongodb.port`.
-//
 // At a minimum, the URL must specify a host, port, and database name. It may
 // omit a username and password if those are provided as
 // `config.mongodb.username` and `config.mongodb.password`, otherwise those
@@ -65,24 +61,9 @@ config.mongodb.options = {
   j: true
 };
 
-// local database config
-//
-// The local database is an optimization for nodes that are able to store
-// local state. It reduces the need to allocate global ids for each new
-// process. If replica sets are used for the main database, the local
-// database should be configured as a separate database instance on each
-// node.
-config.mongodb.local = {};
-config.mongodb.local.enable = false;
-// config.mongodb.local.url = 'mongodb://localhost:27017/local';
-config.mongodb.local.name = 'local';
-config.mongodb.local.collection = 'bedrock_dev';
-config.mongodb.local.writeOptions = {
-  w: 1,
-  j: true,
-  multi: true
-};
-
 config.mongodb.requirements = {};
 // server version requirement with server-style string
 config.mongodb.requirements.serverVersion = '>=4.2';
+
+// this is used by _createUser to add a user as an admin
+//config.mongodb.collection = 'admin-collection';
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,20 +1,10 @@
 /*
  * Bedrock mongodb module.
  *
- * This modules exposes an API for accessing the shared and local databases.
+ * This modules exposes an API for accessing MongoDB databases.
  * This API is mostly used to communicate with the shared database -- this
  * database can be sharded and replicated across multiple machines. Any shared
  * collections are exposed via this module's 'collections' property.
- *
- * The API also exposes a single document in a local database. This database
- * is not sharded or replicated to other machines. It has a single collection
- * with a single document that can be updated atomically. The expectation is
- * that very little data needs to be stored locally (eg: local parts of
- * distributed IDs, etc.). This module exposes the local collection via
- * the 'localCollection' property. The single local document in that
- * collection has two properties: 'id' and 'local'. The value of 'id'
- * is exposed by this module as 'localDocumentId'. The value of 'local' is
- * a JSON object where local properties should be stored.
  *
  * Copyright (c) 2012-2018 Digital Bazaar, Inc. All rights reserved.
  */
@@ -58,7 +48,6 @@ let testMode = false;
 
 // database client(s)
 api.client = null;
-api.localClient = null;
 
 // database itself
 api.db = null;
@@ -66,14 +55,8 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
-// local collection
-api.localCollection = null;
-// local document ID
-api.localDocumentId = 'local';
-
 // default database write options
 api.writeOptions = bedrock.config.mongodb.writeOptions;
-api.localWriteOptions = bedrock.config.mongodb.local.writeOptions;
 
 // load test config
 bedrock.events.on('bedrock.test.configure', () => require('./test.config'));
@@ -115,17 +98,6 @@ function init(callback) {
         init: false
       }, callback);
     }],
-    openLocal: ['initDatabase', (results, callback) => {
-      if(!config.local.enable) {
-        logger.debug('opening local database disabled');
-        return callback(null, null);
-      }
-      logger.info('opening local database', {url: config.local.url});
-      _openDatabase({
-        url: config.local.url,
-        init: false
-      }, callback);
-    }],
     dropCollections: ['open', (results, callback) => {
       api.client = results.open.client;
       api.db = results.open.db;
@@ -142,14 +114,6 @@ function init(callback) {
         fields: {namespace: 1},
         options: {unique: true, background: true}
       }], callback)],
-    setupLocal: ['openLocal', (results, callback) => {
-      // setup machine-local (non-replicated) database
-      api.localClient = results.openLocal;
-      if(!results.openLocal) {
-        return callback();
-      }
-      _setupLocalDatabase(callback);
-    }]
   }, err => {
     if(err) {
       logger.error('could not initialize database', err);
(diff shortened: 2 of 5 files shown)


lib/index.js
```diff
@@ -816,12 +668,12 @@ function _getAdminCredentials(callback) {
...
 function _getAddUserOptions() {
-  const local = bedrock.config.mongodb.local;
+  const config = bedrock.config.mongodb;
   return bedrock.util.extend({}, bedrock.config.mongodb.writeOptions, {
     roles: [
       'dbOwner',
-      {role: 'dbAdmin', db: local.name, collection: local.collection},
-      {role: 'readWrite', db: local.name, collection: local.collection}
+      {role: 'dbAdmin', db: config.name, collection: config.collection},
+      {role: 'readWrite', db: config.name, collection: config.collection}
```
mattcollier, line 676: I believe this pertains to your comments in the description of the PR.  This helper function formerly only transacted with the local database which is not a thing anymore.  Is there a reason why the function would now be altering the regular, non-local database?  Seems to me that this helper function just goes away.

aljones15: just the helper or `_createUser` entirely?

dlongley:
> Also I believe that the user for mongodb is set either in a docker or kubernetes file now possibly via helm chart meaning that _createUser probably should be dropped.

Whatever we do here, we can't have things stop working for dev set ups that don't involve k8s/helm charts.

lib/index.js
davidlehn: I think there is a cascade of code that gets removed with the local db.  The call to _createUser uses db param that is no longer set, so that function probably goes away too.  If that gets removed, then _getAdminCredentials is gone, and the prompt dep, and _getAddUserOptions.  Maybe other code too.

aljones15: @davidlehn cascade of code removed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/52/commits/7fa8314db345860c3e1441673691666fd252f6bd @mattcollier removed `_createUser` and related code.

aljones15:
> > Also I believe that the user for mongodb is set either in a docker or kubernetes file now possibly via helm chart meaning that _createUser probably should be dropped.
> 
> Whatever we do here, we can't have things stop working for dev set ups that don't involve k8s/helm charts.

should not break any functionality for them, but they will need to create a user using the mongodb shell for their `mongod.conf` file. Should I add a readme section for that with links to the MongoDB documentation? Removing _createUser does remove some functionality, but I don't think it's unreasonable to assume a developer can create a mongodb user themself and also we were showing a command prompt, now we throw an error. I think the error is better in most situations such as AWS cloudwatch logs, docker logs, etc.

lib/index.js
```diff
@@ -529,56 +443,31 @@ function _initDatabase(callback) {
...
+      // if the database did not open then the mongoDB server is running,
+      // but the user supplied incorrect information in the config.
+      const {config} = bedrock;
+      const {name, username, connectOptions = {}} = config.mongodb || {};
+      throw new Error('\nThe database "' + name +
+        '" failed to open. Please ensure the database ' + name +
+        ' exists, the primary user "' + username + '" ' +
+        'exists and has sufficient access privileges,' +
+        ' and that "connectOptions.authSource" ' +
+        connectOptions.authSource + ' is set correctly.', results);
```
davidlehn, line 460: - Use BedrockError.
- Is `results` all that useful here?
- If constructing a string like this, maybe use backtick interpolated style: \`foo ${name} bar baz\`.
- However, if using a BedrockError could use a simpler English string and put param in a details object.

davidlehn:
Add CHANGELOG entry with details.  Breaking, local db removed, maybe some user creating issues, etc.

dlongley:
>  but I don't think it's unreasonable to assume a developer can create a mongodb user themself

-1 if a developer didn't need to do this before, then it is unreasonable to add this friction to their process now. It is even more unreasonable if this has to be done for every application/database created. I'm very concerned when we say "it's not unreasonable to have developers do more work." No, it is. One of bedrock's core design goals is to reduce extra processes that create friction for developers in an effort to get to "zero config"/"zero boilerplate" as much as possible. Other frameworks are a pain to use because they have so much additional setup and documentation that must be read. The biggest problems with bedrock all revolve around extra documentation that must be read or doesn't exist -- and this direction would just create more of that.

If the above changes make it so we need extra docs or the developer has to do more things, we must not do it. We must find another way forward.

lib/index.js
aljones15: marking this as resolved as this code is not in the project anymore. You are correct it should have been a `BedrockError` and I'm sorry for not using one.

lib/config.js
```diff
@@ -65,24 +63,9 @@ config.mongodb.options = {
...
-  j: true,
-  multi: true
-};
-
 config.mongodb.requirements = {};
 // server version requirement with server-style string
 config.mongodb.requirements.serverVersion = '>=4.2';
+
+// this is used by _createUser to add a user as an admin
+//config.mongodb.collection = 'admin-collection';
```
JSAssassin, line 71: space after //

aljones15: thanks https://github.com/digitalbazaar/bedrock-mongodb/pull/52/commits/b0fef830eda6826f392124377d2d4b3cff45c318

lib/index.js
```diff
@@ -633,24 +522,30 @@ function _openDatabase(options, callback) {
...
+        auth = {
+          user: config.username,
+          password: config.password
+        };
+      }
+      // authSource should be set in connectOptions
       let opts = {...config.authentication, ...config.connectOptions};
-      const server = new Server(config.host, config.port, opts);
+      let url = config.url;
+      // if the user specified a connection string use it
```
dlongley, line 536:
```suggestion
      // if the user specified a connection URL use it
```

lib/index.js
```diff
@@ -633,24 +522,30 @@ function _openDatabase(options, callback) {
...
+        url = _createUrl(config);
+      }
       if(_usesRoles(results.serverInfo)) {
-        auth.authdb = config.name;
         // authenticate against shared db
-        opts.authSource = config.connectOptions.authSource || config.name;
-        return _loginUser({server, opts, auth, callback});
+        return _loginUser({url, opts, auth, callback});
       }
       // FIXME this code will never be called because serverVersion is 4.2
```
dlongley, line 544: If this is true, can we remove the FIXME or should we throw an error here or something? I'm confused but this.

lib/index.js
```diff
@@ -706,89 +601,68 @@ function _connect(options, callback) {
  *
  * @param {object} options - Options to use.
  * @param {object} options.auth - user and password credentials.
+ * @param {string} options.auth.user - The MongoDB usename.
```
dlongley, line 604:
```suggestion
 * @param {string} options.auth.user - The MongoDB username.
```

lib/index.js
```diff
@@ -706,89 +601,68 @@ function _connect(options, callback) {
...
  *
  * @returns {Promise} The result of the connect.
 */
-async function _loginUser({auth, opts, callback, server}) {
-  const client = new MongoClient(server, {
-    auth,
-    ...opts
-  });
-  return client.connect(callback);
+async function _loginUser({auth, opts, url, callback}) {
```
dlongley, line 612: Should this be `async`? It takes a callback ... and doesn't seem to be used in some backwards compatible way, it only seems to use the `callback` version. Maybe this is ok as is, just asking.

lib/index.js
```diff
@@ -90,19 +73,25 @@ bedrock.events.on('bedrock-cli.ready', () => {
...
+function _createUrl(config) {
+  let url = 'mongodb://';
+  if(config.username) {
+    url += `${config.username}:${config.password}@`;
+  }
+  url += `${config.host}:${config.port}/${config.name}`;
+  // this needs to come last
+  if(config.username) {
+    url +=
+      `?authSource=${config.connectOptions.authSource || 'admin'}`;
```
mattcollier, line 85: Should we just default `connectionOptions.authSource` to `admin` and not have this conditional?

aljones15: We would be specifying auth credentials even when the user is not necessarily using auth. This has led to problems in the past such as setting `username` and `password` in `config.js`. I think it's better to ensure the user sets their `authSource` themselves especially as `admin` is usually the `authSource` only in development. In production you usually have a dedicated database with a different name that holds your mongo users. So this code only runs if a `username` was specified and allows developers running locally to be lazy and not necessarily specify the `authSource`. So I guess in short it is better if `authSource` is set to default in general, if there is a username then ok we need to set it.

lib/index.js
aljones15: I believe when I started I had intended to make this module `async` entirely. As written this is not async so I should probably remove the `async` tag for now. I do think in the future this module should get an over haul to `async` `await` code.

lib/index.js
aljones15: Originally it worked like this: we check the mongo server version. If the version was less than 3.0 I think, we used an older login method. Because the `serverVersion` in `config.js` is set to 4.2 this project now throws if you attempt to connect to a MongoDB database that is lower than 4.2. So some options:

1. just remove the fixme and the server check and just use one login
2. keep the check, but throw if the serverVersion is somehow pre-roles (I forget the exact version of mongodb, but something like 3.0 I think is when roles came into MongoDB).

I had a number three, but I forgot about it.

dlongley: We should get some kind of error if we're connecting to an incompatible database. If that doesn't happen with the driver automatically, we should throw here.

lib/index.js
davidlehn: Doesn't look like it should have async here.  We'll fix up everything at once to be async/await.

aljones15: fixed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/52/commits/056abb3862e88f5d2b28dddf68591cf6124db0ba

lib/index.js
aljones15: now throwing: https://github.com/digitalbazaar/bedrock-mongodb/pull/52/commits/f62a9ad87244f7466f3b22865ce59cbca4fd119f

dlongley:
@davidlehn can you look here? I think everyone else has approved at this point.

aljones15:
@dlongley @mattcollier @davidlehn just so you know I tried this functionality in `bedrock-mongodb ^7.0.0` and currently `adminPrompt: true` with a user and password specified causes bedrock-mongodb to fail. In other words we need to release this as a bug fix for the current version of `bedrock-mongodb`. While it does remove localCollections it does it in a way that should for the most part (there are removed options) be backwards compatible with the original feature.

PR #57, opened 2020-11-04
Use database sepecified in the connect string when calling `client.db()`.

mattcollier:
This in service of specifying ONLY the `url` when connecting.  The operative bit here is the removal of `config.name` from the `client.db()` call that was on L592, and is now on L596. The unexpected behavior caused by this has been haunting me for a while.  The unexpected behavior occurs when one is specifying a self contained url string that is designed to provide the database, username and password, whereby none of these values are set to non-default values in the config.  

For instance: `mongodb://my-user:my-password@mongos.example.com:27075/my_database?ssl=true`.  With this connect string solely, a valid authenticated database connection can be made.  However, with line `client.db(config.name)`, the client tries to connect to a non-existent database, `bedrock_dev` (the default value) and an authentication error ensues because the user that has authenticated to the database `my_database` has no permissions for the `bedrock_dev` database.

`client.db()` when called with no options, returns the database that was from the connect string, and therefore it is redundant at best, and creates a problem if `config.name` is not being set because the `url` config is in effect.

See: https://mongodb.github.io/node-mongodb-native/3.5/api/MongoClient.html#db

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -583,15 +583,19 @@ function _connect(options, callback) {
     // this is usually `admin` in dev and a specific db in production
     connectOptions.authSource = config.connectOptions.authSource || config.name;
   }
+
   MongoClient.connect(options.url, connectOptions, (err, client) => {
-    if(!err && !options.init) {
+    if(err) {
+      return callback(err);
+    }
+
+    if(!options.init) {
       logger.info('connecting to database: ' + options.url);
     }
-    let db = null;
-    if(client) {
-      db = client.db(config.name);
-    }
-    callback(err, {db, client});
+
+    const db = client.db();
+
+    callback(null, {db, client});
   });
 }
 


mattcollier:
I forgot to add that this is simply a patch release as it does not impact typical dev use cases where the typical bedrock.mongodb config options are utilized.  These are composed into the connect string, and the default database returned by `client.db()` is still the right database.

lib/index.js
```diff
@@ -583,15 +583,19 @@ function _connect(options, callback) {
     // this is usually `admin` in dev and a specific db in production
     connectOptions.authSource = config.connectOptions.authSource || config.name;
   }
+
   MongoClient.connect(options.url, connectOptions, (err, client) => {
-    if(!err && !options.init) {
-      logger.info('connecting to database: ' + options.url);
+    if(err) {
+      return callback(err);
```
aljones15, line 589: this great because it lessens the need for a check

lib/index.js
```diff
@@ -583,15 +583,19 @@ function _connect(options, callback) {
...
-    let db = null;
-    if(client) {
-      db = client.db(config.name);
+
+    if(!options.init) {
+      logger.info('connecting to database: ' + options.url);
     }
-    callback(err, {db, client});
+
+    const db = client.db();
```
aljones15, line 596: Ok so you did remove the client check in favor of returning an error callback early, but has this been tested with a failed auth connection to a database? I'm pretty confident it will work (possibly better), but if a connect fails there is no client, there should be an error, but I would like to know what happens if that is the case for sure. This should be pretty simple to do just enableAuth and set a bogus username and password.

aljones15 (changes requested):
I like the changes just reassure me that:

1. This branch has been used in another project
2. This code still throws if there is an authentication error

mattcollier:
@aljones15 this branch has now been tested using authentication based solely on the URL config option (user provided connection string), specifying username/password via regular bedrock config (auto-composed connect string) and no authentication (test suite).

This does not alter any existing error handling.

aljones15 (approved):
Sorry about the hold up and thanks for the tests.
I'm approving this.

Issue #58, opened 2020-11-05
Adjust logging for config.url


mattcollier:
There are many instances where `config.url` is being logged.  This url will often include login credentials in the form:
```
mongodb://my-user:my-password@mongos.example.com:27075/my_database?ssl=true
```
Login credentials should not be surfaced in the logs.  This needs to be updated immediately and therefore I propose that all logging of `config.url` be removed entirely. Perhaps in the future a more sophisticated logging method that redacts the credentials can be developed. 

See: https://github.com/digitalbazaar/bedrock-mongodb/blob/master/lib/index.js#L101

mattcollier:
Work on this should be based on the branch associated with this PR: https://github.com/digitalbazaar/bedrock-mongodb/pull/57

dmitrizagidulin:
+1.

PR #59, opened 2020-11-05
Sanitize `config.url`

JSAssassin:
addresses #58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # bedrock-mongodb ChangeLog
 
-## 8.0.4 - 2020-11-05
+## 8.0.4 - 2020-11-TBD
 
 ### Fixed
 - Log sanitized `config.url`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # bedrock-mongodb ChangeLog
 
+## 8.0.4 - 2020-11-05
+
+### Fixed
+- Log sanitized `config.url`.
+
 ## 8.0.3 - 2020-11-05
 
 ### Fixed
--- a/lib/index.js
+++ b/lib/index.js
@@ -89,18 +89,7 @@ function _createUrl(config) {
 
 function _sanitizeUrl(url) {
   const urlParts = require('url').parse(url);
-  let user;
-  if(urlParts.auth) {
-    user = urlParts.auth.split(':')[0];
-  }
-  let sanitizedUrl;
-  if(user) {
-    sanitizedUrl =
-      `${urlParts.protocol}//${user}:****@${urlParts.host}${urlParts.path}`;
-  } else {
-    sanitizedUrl = url;
-  }
-  return sanitizedUrl;
+  return `${urlParts.protocol}//${urlParts.host}${urlParts.path}`;
 }
 
 function init(callback) {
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -87,6 +87,22 @@ function _createUrl(config) {
...
+function _sanitizeUrl(url) {
+  const urlParts = require('url').parse(url);
+  let user;
+  if(urlParts.auth) {
+    user = urlParts.auth.split(':')[0];
+  }
+  let sanitizedUrl;
+  if(user) {
+    sanitizedUrl =
+      `${urlParts.protocol}//${user}:****@${urlParts.host}${urlParts.path}`;
```
mattcollier, line 99: my preference would be that we don't even log the user, @JSAssassin @aljones15 @davidlehn @dlongley @gannan08 thoughts?

mattcollier: Which would also simply the implementation somewhat here.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # bedrock-mongodb ChangeLog
 
+## 8.0.4 - 2020-11-05
```
mattcollier, line 3: date needs to be TBD or similar.

lib/index.js
gannan08: In a production deployment, I do not see the added value for logging the user.

JSAssassin: @mattcollier @gannan08 agreed. I can update that.

lib/index.js
```diff
@@ -87,6 +87,11 @@ function _createUrl(config) {
   return url;
 }
 
+function _sanitizeUrl(url) {
```
mattcollier, line 90: Please move helper to bottom of the file along with any other helpers (if any).  Private functions like this are hoisted and available before they are defined.

JSAssassin: Fixed - df652a8

lib/index.js
```diff
@@ -745,3 +731,22 @@ function _dropCollections(callback) {
...
+  // this needs to come last
+  if(config.username) {
+    url +=
+      `?authSource=${config.connectOptions.authSource || 'admin'}`;
+  }
+  return url;
+}
+
+function _sanitizeUrl(url) {
+  const urlParts = require('url').parse(url);
```
mattcollier, line 750: Looking forward to when this and every other lib is converted to ESM syntax.  Please do not add CommonJS require into functions like this.  Pull in `url` at the top of the file along with all other deps which will make it easier to convert later.

lib/index.js
davidlehn: NODE_ENV values could also be used to control this if it turns out more debug info would be useful.

mattcollier: -1 to bringing NODE_ENV into this, at least at this time.

lib/index.js
JSAssassin: Updated - 08b9847

mattcollier (approved):

davidlehn (approved):

davidlehn (changes requested):
I think there are a few more places to change.
- The errors also include config.url.  If those trigger, logged error messages would have full URL.
- _connect is logging url twice.  At the top and in the MongoClient.connect callback. (same message, which seems confusing)

davidlehn (approved):

Issue #64, opened 2021-02-05
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.


aljones15:
This a non-fatal error that occurred in a project using this library.
We need to look into why these are options are being passed in provided this is definitely a mongodb error.

Error occurred on a mac powerbook and platform could be a factor.

Mongodb docs for write options

[fsync](https://docs.mongodb.com/manual/reference/command/fsync/)
[writeConcern (includes w, j, & wtimeout](https://docs.mongodb.com/manual/reference/write-concern/)

this came in with [`node native mongodb 3.6.4`](https://github.com/mongodb/node-mongodb-native/blob/3.6/HISTORY.md) 
    * Deprecate top-level write concern option keys  ([#2624](https://github.com/mongodb/node-mongodb-native/issues/2624)) ([0516d93](https://github.com/mongodb/node-mongodb-native/commit/0516d93f74de4b58a99     e8455e59678d4b09cd4a7))

aljones15:
if anyone needs to replicate this, just running [this test suite](https://github.com/digitalbazaar/bedrock-edv-storage/tree/master/test) generates the warning:

```js
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
```

mattcollier:
@aljones15 thanks for writing this up.  I just saw this for the first time myself on that `bedrock-edv-storage` test suite you referenced.  This would mean it's not platform specific.

aljones15:
> @aljones15 thanks for writing this up. I just saw this for the first time myself on that `bedrock-edv-storage` test suite you referenced. This would mean it's not platform specific.

can you specify the platform and perhaps mongodb server version? I've only seen it on mac.

mattcollier:
Errors can be seen on CI: https://github.com/digitalbazaar/bedrock-edv-storage/runs/1856141883?check_suite_focus=true

aljones15:
> Errors can be seen on CI: https://github.com/digitalbazaar/bedrock-edv-storage/runs/1856141883?check_suite_focus=true

thank you.

aljones15:
as of `mongodb 3.6.4` many of the node driver's write operations have defaults that do not use `writeConcern`. The following is from mongodb's own test suite (not ours):

```js
  Write Concern
    ✓ should respect writeConcern from uri (85ms)
    spec tests
      default-write-concern-2.6 - replicaset,sharded,single
[...]

PR #65, opened 2021-02-08
Use write concern

aljones15:
Updates our settings so we do not have deprecated uses of `w`, `j`, `wtimeout`, and `fsync`.

Those options now need to be inside of a `writeConcern` object.

This addresses:

https://github.com/digitalbazaar/bedrock-mongodb/issues/64

This is a draft PR until it can be confirmed it removes this deprecation error:

```js
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
```

in multiple projects.

Removes deprecation warning in:

1. bedrock-mongodb
2. bedrock-edv-storage
3. bedrock-account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # bedrock-mongodb ChangeLog
 
+## 8.2.0 -
+
+### Removed
+- `options.mongodb`
+- Calls on `w`, `j`, `wtimeout`, and `fsync`.
+
+### Added
+- Support for mongodb's new `writeConcern` option.
+
 ## 8.1.1 - 2020-12-15
 
 ### Fixed
--- a/lib/config.js
+++ b/lib/config.js
@@ -41,8 +41,10 @@ config.mongodb.authentication = {
   // authMechanism: 'MONGODB-CR'
 };
 config.mongodb.connectOptions = {
-  w: 'majority',
-  j: true,
+  writeConcern: {
+    w: 'majority',
+    j: true
+  },
   useUnifiedTopology: true,
   serverSelectionTimeoutMS: 30000,
   autoReconnect: false,
@@ -53,8 +55,10 @@ config.mongodb.connectOptions = {
   promoteBuffers: true,
 };
 config.mongodb.writeOptions = {
-  j: true,
-  w: 'majority',
+  writeConcern: {
+    j: true,
+    w: 'majority'
+  },
   // as mongo 4.2 this is no longer used by updateOne or updateMany
   // FIXME remove this once all dependencies have been upgraded
   multi: true,
@@ -63,9 +67,11 @@ config.mongodb.writeOptions = {
 };
 /* DEPRECATED */
 config.mongodb.options = {
-  w: 'majority',
-  journal: true,
-  j: true
+  writeConcern: {
+    w: 'majority',
+    j: true
+  },
+  journal: true
 };
 
 config.mongodb.requirements = {};
--- a/lib/config.js
+++ b/lib/config.js
@@ -59,9 +59,6 @@ config.mongodb.writeOptions = {
     j: true,
     w: 'majority'
   },
-  // as mongo 4.2 this is no longer used by updateOne or updateMany
-  // FIXME remove this once all dependencies have been upgraded
-  multi: true,
   // this is used by insert methods
   forceServerObjectId: true,
 };
@@ -70,8 +67,7 @@ config.mongodb.options = {
   writeConcern: {
     w: 'majority',
     j: true
-  },
-  journal: true
+  }
 };
 
 config.mongodb.requirements = {};
--- a/lib/config.js
+++ b/lib/config.js
@@ -43,7 +43,7 @@ config.mongodb.authentication = {
 
 const writeConcern = config.mongodb.writeConcern = {
   w: 'majority',
-  j: true
+  j: true,
 };
 
 config.mongodb.connectOptions = {
--- a/lib/config.js
+++ b/lib/config.js
@@ -41,14 +41,8 @@ config.mongodb.authentication = {
   // authMechanism: 'MONGODB-CR'
 };
 
-const writeConcern = config.mongodb.writeConcern = {
-  w: 'majority',
-  j: true,
-};
-
 // this is used when making an initial connection
 config.mongodb.connectOptions = {
-  writeConcern,
   useUnifiedTopology: true,
   serverSelectionTimeoutMS: 30000,
   autoReconnect: false,
@@ -61,7 +55,10 @@ config.mongodb.connectOptions = {
 
 // this is used by insert methods
 config.mongodb.writeOptions = {
-  writeConcern,
+  writeConcern: {
+    w: 'majority',
+    j: true,
+  },
   forceServerObjectId: true,
 };
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -62,10 +62,6 @@ config.mongodb.writeOptions = {
   // this is used by insert methods
   forceServerObjectId: true,
 };
-/* DEPRECATED */
-config.mongodb.options = {
-  writeConcern
-};
 
 config.mongodb.requirements = {};
 // server version requirement with server-style string
--- a/lib/config.js
+++ b/lib/config.js
@@ -41,7 +41,7 @@ config.mongodb.authentication = {
   // authMechanism: 'MONGODB-CR'
 };
 
-// this is used when making an initial connection
+// this is used when making connections to the database
 config.mongodb.connectOptions = {
   useUnifiedTopology: true,
   serverSelectionTimeoutMS: 30000,
--- a/lib/config.js
+++ b/lib/config.js
@@ -46,6 +46,7 @@ const writeConcern = config.mongodb.writeConcern = {
   j: true,
 };
 
+// this is used when making an initial connection
 config.mongodb.connectOptions = {
   writeConcern,
   useUnifiedTopology: true,
@@ -57,9 +58,10 @@ config.mongodb.connectOptions = {
   // promotes binary BSON values to native Node.js buffers
   promoteBuffers: true,
 };
+
+// this is used by insert methods
 config.mongodb.writeOptions = {
   writeConcern,
-  // this is used by insert methods
   forceServerObjectId: true,
 };
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -40,11 +40,14 @@ config.mongodb.authentication = {
   // used by MongoDB < 3.0
   // authMechanism: 'MONGODB-CR'
 };
+
+const writeConcern = config.mongodb.writeConcern = {
+  w: 'majority',
+  j: true
+};
+
 config.mongodb.connectOptions = {
-  writeConcern: {
-    w: 'majority',
-    j: true
-  },
+  writeConcern,
   useUnifiedTopology: true,
   serverSelectionTimeoutMS: 30000,
   autoReconnect: false,
@@ -55,19 +58,13 @@ config.mongodb.connectOptions = {
   promoteBuffers: true,
 };
 config.mongodb.writeOptions = {
-  writeConcern: {
-    j: true,
-    w: 'majority'
-  },
+  writeConcern,
   // this is used by insert methods
   forceServerObjectId: true,
 };
 /* DEPRECATED */
 config.mongodb.options = {
-  writeConcern: {
-    w: 'majority',
-    j: true
-  }
+  writeConcern
 };
 
 config.mongodb.requirements = {};
--- a/lib/config.js
+++ b/lib/config.js
@@ -53,7 +53,7 @@ config.mongodb.connectOptions = {
   promoteBuffers: true,
 };
 
-// this is used by insert methods
+// this is used when writing to the database
 config.mongodb.writeOptions = {
   writeConcern: {
     w: 'majority',
--- a/lib/index.js
+++ b/lib/index.js
@@ -131,6 +131,7 @@ function init(callback) {
 api.openCollections = brCallbackify(async names => {
   // remove collections that are already open
   const unopened = [];
+  const {writeConcern} = bedrock.config.mongodb;
   for(const name of names) {
     if(!(name in api.collections)) {
       unopened.push(name);
@@ -146,7 +147,7 @@ api.openCollections = brCallbackify(async names => {
   await Promise.all(unopened.map(async name => {
     logger.debug('creating collection: ' + name);
     try {
-      await api.db.createCollection(name);
+      await api.db.createCollection(name, {writeConcern});
     } catch(e) {
       if(!api.isAlreadyExistsError(e)) {
         throw e;
@@ -159,7 +160,7 @@ api.openCollections = brCallbackify(async names => {
   logger.debug('opening collections', {collections: unopened});
   const collections = {};
   await Promise.all(unopened.map(async name => {
-    collections[name] = await api.db.collection(name);
+    collections[name] = await api.db.collection(name, {writeConcern});
   }));
 
   // merge results into collection cache
--- a/lib/index.js
+++ b/lib/index.js
@@ -288,7 +288,7 @@ api.createIndexes = brCallbackify(async options => {
  * @return the new GridFSBucket instance
  */
 api.createGridFSBucket = options => {
-  const opts = {...options, ...api.writeOptions};
+  const opts = {...bedrock.config.mongodb.writeOptions, ...options};
   return new mongo.GridFSBucket(api.db, opts);
 };
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -554,7 +554,6 @@ function _connect(options, callback) {
   }
   const {writeConcern} = config.writeOptions;
   let connectOptions = {...config.connectOptions, writeConcern};
-
   // convert legacy connect options
   if('socketOptions' in config.connectOptions) {
     connectOptions = {
--- a/lib/index.js
+++ b/lib/index.js
@@ -58,8 +58,8 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
-const {writeConcern} = bedrock.config.mongodb.writeOptions;
-
+// Note: api.writeOptions is deprecated
+// and will be removed in release 9.0
 // default database write options
 api.writeOptions = {...bedrock.config.mongodb.writeOptions};
 
@@ -163,7 +163,8 @@ api.openCollections = brCallbackify(async names => {
   const {writeConcern} = api.db.options;
   await Promise.all(unopened.map(async name => {
     // Note: We only pass `{writeConcern}` here to get around a bug in mongodb
-    // node driver 3.6.4 where `writeConcern` from `db` is not passed to collection
+    // node driver 3.6.4 where `writeConcern` from `db` is not passed to
+    // collection
     collections[name] = await api.db.collection(name, {writeConcern});
   }));
 
@@ -287,7 +288,7 @@ api.createIndexes = brCallbackify(async options => {
  * @return the new GridFSBucket instance
  */
 api.createGridFSBucket = options => {
-  const opts = Object.assign({}, {...api.writeOptions}, options);
+  const opts = {...options, ...api.writeOptions};
   return new mongo.GridFSBucket(api.db, opts);
 };
 
@@ -512,11 +513,12 @@ function _openDatabase(options, callback) {
           password: config.password
         };
       }
+
       // authSource should be set in connectOptions
       const opts = {
         ...config.authentication,
         ...config.connectOptions,
-        writeConcern
+        ...config.writeOptions,
       };
       let url = config.url;
       // if the user specified a connection URL use it
@@ -550,7 +552,7 @@ function _connect(options, callback) {
   if(!options.init) {
     logger.info('connecting to database: ' + _sanitizeUrl(options.url));
   }
-
+  const {writeConcern} = config.writeOptions;
   let connectOptions = {...config.connectOptions, writeConcern};
 
   // convert legacy connect options
--- a/lib/index.js
+++ b/lib/index.js
@@ -61,7 +61,7 @@ api.collections = {};
 const {writeConcern} = bedrock.config.mongodb.writeOptions;
 
 // default database write options
-api.writeOptions = {...bedrock.config.mongodb.writeOptions, writeConcern};
+api.writeOptions = {...bedrock.config.mongodb.writeOptions};
 
 // load test config
 bedrock.events.on('bedrock.test.configure', () => require('./test.config'));
@@ -162,6 +162,8 @@ api.openCollections = brCallbackify(async names => {
   const collections = {};
   const {writeConcern} = api.db.options;
   await Promise.all(unopened.map(async name => {
+    // FIXME: this is to get around a bug in mongodb node driver 3.6.4
+    // where the writeConcern from db is not being passed to collection
     collections[name] = await api.db.collection(name, {writeConcern});
   }));
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -61,7 +61,7 @@ api.collections = {};
 // Note: api.writeOptions is deprecated
 // and will be removed in release 9.0
 // default database write options
-api.writeOptions = {...bedrock.config.mongodb.writeOptions};
+api.writeOptions = bedrock.config.mongodb.writeOptions;
 
 // load test config
 bedrock.events.on('bedrock.test.configure', () => require('./test.config'));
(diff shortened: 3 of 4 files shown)


lib/index.js
```diff
@@ -146,7 +147,7 @@ api.openCollections = brCallbackify(async names => {
   await Promise.all(unopened.map(async name => {
     logger.debug('creating collection: ' + name);
     try {
-      await api.db.createCollection(name);
+      await api.db.createCollection(name, {writeConcern});
```
mattcollier, line 150: Since the `writeConcern` is passed into the client on initialization in `connectionOptions` is this also required?

lib/config.js
```diff
@@ -40,9 +40,14 @@ config.mongodb.authentication = {
   // used by MongoDB < 3.0
   // authMechanism: 'MONGODB-CR'
 };
-config.mongodb.connectOptions = {
+
+const writeConcern = config.mongodb.writeConcern = {
   w: 'majority',
-  j: true,
+  j: true
```
mattcollier, line 46: I'm a fan of the dangling comma, so I would prefer not seeing them removed.  They are legal based on current linting rules so I don't see any reason it would need to be removed.

lib/config.js
```diff
@@ -53,19 +58,13 @@ config.mongodb.connectOptions = {
   promoteBuffers: true,
 };
 config.mongodb.writeOptions = {
-  j: true,
-  w: 'majority',
-  // as mongo 4.2 this is no longer used by updateOne or updateMany
-  // FIXME remove this once all dependencies have been upgraded
-  multi: true,
+  writeConcern,
```
dlongley, line 61: If this is already set on `connectOptions` -- does it need to be set here as well? Or will `connectOptions` form the default that is used and we can omit it here?

lib/config.js
```diff
@@ -40,9 +40,14 @@ config.mongodb.authentication = {
   // used by MongoDB < 3.0
   // authMechanism: 'MONGODB-CR'
 };
-config.mongodb.connectOptions = {
+
+const writeConcern = config.mongodb.writeConcern = {
```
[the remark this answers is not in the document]
aljones15: This pattern does consolidate `writeConcern` in one place. We re-use `config.mongodb.writeConcern` everywhere we had those parameters before.

lib/index.js
aljones15: it is actually not needed here, it is needed here though:
https://github.com/digitalbazaar/bedrock-mongodb/blob/31a3c1d8a39a6655735d537382f4a3f68da581d5/lib/index.js#L162-L164

lib/config.js
aljones15: This is a change from previous linting rules. When I joined DB a previous PR was held up do to a dangling comma. I have no issue with them either so I will re-add it. We might want to make a lint rule for it: https://eslint.org/docs/rules/comma-dangle

I added an issue here:
https://github.com/digitalbazaar/eslint-config-digitalbazaar/issues/49

lib/config.js
```diff
@@ -53,19 +58,13 @@ config.mongodb.connectOptions = {
...
-  w: 'majority',
-  // as mongo 4.2 this is no longer used by updateOne or updateMany
-  // FIXME remove this once all dependencies have been upgraded
-  multi: true,
+  writeConcern,
   // this is used by insert methods
   forceServerObjectId: true,
 };
 /* DEPRECATED */
 config.mongodb.options = {
```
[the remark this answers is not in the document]
aljones15: has been deprecated since version `2.0.0` so probably safe to remove.

lib/config.js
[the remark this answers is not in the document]
dlongley: I see them being used in `config.mongodb.connectOptions`, which could be edited independently from `config.mongodb.writeConcern`, resulting in the problem I'm worried about. I'm not talking about them being in the same variable in a file, I'm talking about exposing multiple interfaces (`config.mongodb.connectOptions`, `bedrock.mongodb.options`) that all use `writeConcern` but that could get out of sync with one another.

Consider the user that does this in an extending config file:

```js
config.mongodb.connectOptions.writeConcern = {/*something else*/};
```

This will cause `writeConcern` to get out of sync in the other places in a way that shouldn't be possible (devs get burned).

Can we either:

1. Ensure that we're using computed variables in some way such that if they are changed in one place, they are automatically changed in all the places, OR
2. Ensure that the code always "assembles" the config by overlaying `config.mongodb.writeConcern` on top of the other options where used, OR
3. Prevent changing `writeConcern` directly inside of `config.mongodb.connectOptions` (et. al), OR
4. Document the danger of changing it directly, OR
5. Some other option that mitigates the concern raised?

lib/config.js
aljones15: the answer to this is inconsistent at this point. These deprecation notices are brand new and mongodb's native node driver still stumbles on them all over the place. The defaults for all write options in the mongodb node driver as of this morning still used `w`, `j`, `fsync` in options and not in the newer option `writeConcern`. So basically, we can not rely on mongodb to satisfy their own deprecation in the current release of the driver.

lib/config.js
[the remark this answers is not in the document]
dlongley: But do we even use it anywhere? If not, it should definitely be removed. If we're using it, then it will take another major rev if we remove it now -- but we probably should so we can stop maintaining it.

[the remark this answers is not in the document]
aljones15: I am unable to find anywhere in our code base it is used. There could be a developer external to use using it. This is a major oversite though as we are on version 8, and it has been marked deprecated since version 2.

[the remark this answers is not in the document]
dlongley: The options are limited to our usage in this module, so if this module doesn't use them, they can be safely removed. Let's remove them now.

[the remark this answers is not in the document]
aljones15: Removed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/7eeb37cb9491e087f06579ea8ecd1fe80cbfedae

lib/index.js
dlongley: Can you confirm that `https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L436-L437` doesn't automatically merge in the defaults that were set earlier when the client was made?

dlongley: Similarly, it looks like `db.createCollection` follows this code path to merge in the `writeConcern` from `db`:

https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L538

https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/operations/create_collection.js#L27-L29

https://github.com/mongodb/node-mongodb-native/blob/e8ac5585068e8fe897cb302f44a8e64b7fa201cc/lib/operations/command.js#L34

https://github.com/mongodb/node-mongodb-native/blob/e8ac5585068e8fe897cb302f44a8e64b7fa201cc/lib/utils.js#L437

So it seems that adding `writeConcern` in either place shouldn't be needed -- can we confirm that if we don't include it that the options aren't set properly (i.e., that they don't use the options we set on `db`)?

aljones15: can confirm it is needed in the `db.collection` call. Perhaps the code you are referencing is in the next release of the driver?

dlongley: Code is from version 3.6 of the driver:

https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L75 
https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L186

How are you confirming it's needed?

dlongley: Also, you linked to the function call `db()` which would be called on a client instance which would have been created via this: `http://mongodb.github.io/node-mongodb-native/3.6/api/MongoClient.html` (which takes `writeConcern`).

lib/config.js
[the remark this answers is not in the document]
aljones15: This adds a layer of `writeConcern`, but does not freeze `writeConcern` itself: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/686656053ecc976ab6420dad2fe7f9af030106f8

lib/index.js
aljones15: code as is:

```js
Running Mocha tests...

  api
    openCollection
      ✓ should open a collection
      ✓ should open several collections (53ms)
    createIndex
      ✓ should create an index
    collections
      ✓ should have collections
      ✓ should insertOne into a collection
      ✓ should insertMany into a collection
      ✓ should findOne in a collection
      ✓ should find many records in a collection
```

code with:

```js
  await Promise.all(unopened.map(async name => {
    collections[name] = await api.db.collection(name, {writeConcern});
  }));
```

changed to:

```js
  await Promise.all(unopened.map(async name => {
    collections[name] = await api.db.collection(name);
  }));
```

results in:

```js
  api
    openCollection
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
      ✓ should open a collection
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
Top-level use of w, wtimeout, j, and fsync is deprecated. Use writeConcern instead.
      ✓ should open several collections (42ms)
    createIndex
      ✓ should create an index
    collections
      ✓ should have collections
      ✓ should insertOne into a collection
      ✓ should insertMany into a collection
      ✓ should findOne in a collection
      ✓ should find many records in a collection


  8 passing (120ms)
```

also in the issue you can see where mongodb (the company/ maintainer of the driver)'s own test suite is throwing these depreaction notices due to defaults in their write api.

lib/config.js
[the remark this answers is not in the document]
aljones15: This is a proposal for a way of making `writeConcern` harder to mess up using an es6 class and getters and setters: https://gist.github.com/aljones15/a9197fc151d2e8ca39ad29de343c2dde

lib/index.js
```diff
@@ -58,8 +58,10 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
+const {writeConcern} = bedrock.config.mongodb;
```
dlongley, line 61: I just made this comment accidentally on a commit instead of here, but can we eliminate:

`bedrock.config.mongodb.writeConcern` and instead just use `bedrock.config.mongodb.writeOptions.writeConcern` in its place? Then if we layer that on top of the `connectOptions` when we use it, I think it will address my concern.

lib/index.js
dlongley: It looks like `collectionKeys` is missing `writeConcern`:

https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L385-L395

Which means that [mergeOptionsAndWriteConcern](https://github.com/mongodb/node-mongodb-native/blob/3.6/lib/db.js#L437) fails to copy that over. We should comment on having to pass `writeConcern` to `db.collection()` on that basis, and we should use the write concern from the `db` instance, not from the config -- effectively "fixing" the bug in the driver.

lib/index.js
```diff
@@ -146,7 +148,7 @@ api.openCollections = brCallbackify(async names => {
   await Promise.all(unopened.map(async name => {
     logger.debug('creating collection: ' + name);
     try {
-      await api.db.createCollection(name);
+      await api.db.createCollection(name, {writeConcern});
```
dlongley, line 151:
```suggestion
      await api.db.createCollection(name);
```

lib/index.js
```diff
@@ -159,7 +161,7 @@ api.openCollections = brCallbackify(async names => {
   logger.debug('opening collections', {collections: unopened});
   const collections = {};
   await Promise.all(unopened.map(async name => {
```
dlongley, line 163:
```suggestion
  const {writeConcern} = api.db.options;
  await Promise.all(unopened.map(async name => {
```

dlongley: We need to remove the top-level `writeConcern` const that is defined as well.

lib/index.js
dlongley: This line isn't safe -- the config can be changed on a bedrock event. This top-level `writeConcern` should be removed.

dlongley (changes requested):

lib/config.js
```diff
@@ -52,21 +52,15 @@ config.mongodb.connectOptions = {
   // promotes binary BSON values to native Node.js buffers
   promoteBuffers: true,
 };
+
+// this is used by insert methods
```
dlongley, line 56:
```suggestion
// this is used when writing to the database
```

lib/config.js
```diff
@@ -40,9 +40,9 @@ config.mongodb.authentication = {
   // used by MongoDB < 3.0
   // authMechanism: 'MONGODB-CR'
 };
+
+// this is used when making an initial connection
```
dlongley, line 44:
```suggestion
// this is used when making connections to the database
```

lib/index.js
```diff
@@ -58,8 +58,10 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
+const {writeConcern} = bedrock.config.mongodb.writeOptions;
```
dlongley, line 61: This isn't safe to use and should be removed. It can change during bedrock startup.

lib/index.js
```diff
@@ -58,8 +58,10 @@ api.db = null;
...
+
 // default database write options
-api.writeOptions = bedrock.config.mongodb.writeOptions;
+api.writeOptions = {...bedrock.config.mongodb.writeOptions};
```
dlongley, line 64: We should mark this as deprecated and remove it in the next major release.

lib/index.js
```diff
@@ -158,8 +160,11 @@ api.openCollections = brCallbackify(async names => {
   // open the collections
   logger.debug('opening collections', {collections: unopened});
   const collections = {};
+  const {writeConcern} = api.db.options;
   await Promise.all(unopened.map(async name => {
-    collections[name] = await api.db.collection(name);
+    // FIXME: this is to get around a bug in mongodb node driver 3.6.4
+    // where the writeConcern from db is not being passed to collection
```
dlongley, line 166: We try not to add FIXMEs these days and `this` isn't clear to people without the context of this PR.

```suggestion
    // Note: We only pass `{writeConcern}` here to get around a bug in mongodb
    // node driver 3.6.4 where `writeConcern` from `db` is not passed to collection
```

lib/index.js
```diff
@@ -282,7 +287,7 @@ api.createIndexes = brCallbackify(async options => {
  * @return the new GridFSBucket instance
  */
 api.createGridFSBucket = options => {
-  const opts = Object.assign({}, {writeConcern: api.writeOptions}, options);
+  const opts = Object.assign({}, {...api.writeOptions}, options);
```
dlongley, line 290:
```suggestion
  const opts = {...api.writeOptions, ...options);
```

lib/index.js
```diff
@@ -508,7 +513,11 @@ function _openDatabase(options, callback) {
         };
       }
       // authSource should be set in connectOptions
-      const opts = {...config.authentication, ...config.connectOptions};
+      const opts = {
+        ...config.authentication,
+        ...config.connectOptions,
+        writeConcern
```
dlongley, line 519:
```suggestion
        ...config.writeOptions.writeConcern
```

lib/index.js
```diff
@@ -542,7 +551,7 @@ function _connect(options, callback) {
     logger.info('connecting to database: ' + _sanitizeUrl(options.url));
   }
 
-  let connectOptions = {...config.connectOptions};
+  let connectOptions = {...config.connectOptions, writeConcern};
```
dlongley, line 554:
```suggestion
  let connectOptions = {...config.connectOptions, config.writeOptions.writeConcern};
```

dlongley (changes requested):

lib/index.js
aljones15: Addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
aljones15: marked with a note here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
aljones15: I believe we want out `writeOptions` to layer over the other ones correct?

```js
var a = {foo: true}
undefined
var b = {foo: false}
undefined
var c = {...a, ...b}
undefined
c
{foo: false}
```
so I think we want this to be:

```js
{...options, ...api.writeOptions}
```

so that the writeConcern layers over the the one passed into to `createGridFSBucket` or am I not getting something here?

made the changes here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
aljones15: this would result result in the options inside of writeConcern being spread into the options itself. So I just decided to do

```js
...config.writeOptions
```

here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
aljones15: Addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
aljones15: just so this has a comment. This was addressed here: https://github.com/digitalbazaar/bedrock-mongodb/commit/163a951857899ac70ee2944a892eb1b6eda2a484

lib/index.js
dlongley: No, we want the caller of the API to be able to override the defaults that come from `api.writeOptions`.

lib/index.js
```diff
@@ -282,7 +288,7 @@ api.createIndexes = brCallbackify(async options => {
  * @return the new GridFSBucket instance
  */
 api.createGridFSBucket = options => {
-  const opts = Object.assign({}, {writeConcern: api.writeOptions}, options);
+  const opts = {...options, ...api.writeOptions};
```
dlongley, line 291: Also, we should not use `api.writeOptions` since it doesn't get updated when the config changes (and we're deprecating it ... so it shouldn't be used anywhere).

```suggestion
  const opts = {...bedrock.config.mongodb.writeOptions, ...options};
```

lib/index.js
```diff
@@ -58,8 +58,10 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
+// Note: api.writeOptions is deprecated
+// and will be removed in release 9.0
 // default database write options
-api.writeOptions = bedrock.config.mongodb.writeOptions;
+api.writeOptions = {...bedrock.config.mongodb.writeOptions};
```
dlongley, line 64: I think we're going to have to not make this shallow copy -- it will change the API.

```suggestion
api.writeOptions = bedrock.config.mongodb.writeOptions;
```

dlongley (approved):
This LGTM now other than the fact that, unfortunately, it seems like it's a major breaking change.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # bedrock-mongodb ChangeLog
 
+## 8.2.0 -
+
+### Removed
+- `options.mongodb`
+- Calls on `w`, `j`, `wtimeout`, and `fsync`.
```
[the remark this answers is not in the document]
aljones15: that seems to be slightly smarter than what mongodb did which is make this change in a minor update, but as it stands deprecation notice withstanding, existing libraries using this package, could still pass in their `w`, `j`, `wtimeout` etc and just get a deprecation notice. BTW calling `collection.indexExists` also causes the deprecation warning and does not accept a `writeConcern` in the options. So we are waiting on mongodb to fix bugs with their driver in that case.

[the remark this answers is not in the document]
dlongley: >  could still pass in their w, j, wtimeout etc and just get a deprecation notice.

What happens when `writeConcern` *and* the top-level values are passed in? Which gets precedence? If it is `writeConcern`, then it is a breaking change.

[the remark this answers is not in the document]
aljones15: ok so I did this test:

```js
// this is used when making connections to the database
config.mongodb.connectOptions = { 
  useUnifiedTopology: true,
  serverSelectionTimeoutMS: 30000,
  autoReconnect: false,
  useNewUrlParser: true,
  // the db to authenticate against
  authSource: undefined,
  // promotes binary BSON values to native Node.js buffers
  promoteBuffers: true,
  j: false,
  w: 0
};

// this is used when writing to the database
config.mongodb.writeOptions = { 
  writeConcern: {
    w: 'majority',
    j: true,
  },  
  forceServerObjectId: true,
};

// result is
api.db.options {
  authSource: undefined,
  promoteBuffers: true,
  j: false,
  w: 0,
  writeConcern: { j: false, w: 0 },
  retryWrites: true,
  readPreference: ReadPreference { mode: 'primary', tags: undefined, hedge: undefined },
  promiseLibrary: [Function: Promise]
}
```js

so the older top level options seem to take precedence over the options from `writeConcern`.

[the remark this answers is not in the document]
Comment (aljones15) [reply CHANGELOG.md:7]: I run ran it second time after this [commit](https://github.com/digitalbazaar/bedrock-mongodb/pull/65/commits/33c3915449939327274f8bca3b5139f3c41ffda4) which removed some writeOptions set via test.config.js

Before adding top level ops:

```js
// this is used when making connections to the database
config.mongodb.connectOptions = { 
  useUnifiedTopology: true,
  serverSelectionTimeoutMS: 30000,
  autoReconnect: false,
  useNewUrlParser: true,
  // the db to authenticate against
  authSource: undefined,
  // promotes binary BSON values to native Node.js buffers
  promoteBuffers: true,
};

// this is used when writing to the database
config.mongodb.writeOptions = { 
  writeConcern: {
    w: 'majority',
    j: true,
  },  
  forceServerObjectId: true,
};

api.db.options {
  authSource: undefined,
  promoteBuffers: true,
  writeConcern: { w: 'majority', j: true },
  w: 'majority',
  j: true,
  retryWrites: true,
  readPreference: ReadPreference { mode: 'primary', tags: undefined, hedge: undefined },
  promiseLibrary: [Function: Promise]
}
```

After setting top level options:

```js
// this is used when making connections to the database
config.mongodb.connectOptions = { 
  useUnifiedTopology: true,
  serverSelectionTimeoutMS: 30000,
  autoReconnect: false,
  useNewUrlParser: true,
  // the db to authenticate against
  authSource: undefined,
  // promotes binary BSON values to native Node.js buffers
  promoteBuffers: true,
  w: 0,
  j: false
};

// this is used when writing to the database
config.mongodb.writeOptions = { 
  writeConcern: { 
    w: 'majority',
    j: true,
  },
  forceServerObjectId: true,
};

api.db.options {
  authSource: undefined,
  promoteBuffers: true,
  w: 0,
  j: false,
  writeConcern: { w: 0, j: false },
  retryWrites: true,
  readPreference: ReadPreference { mode: 'primary', tags: undefined, hedge: undefined },
  promiseLibrary: [Function: Promise]
}
```

so yeah it does appear it turns the top level options into writeConcern when both are defined and passed in.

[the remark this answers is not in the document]
dlongley: @aljones15 -- Ok, given these test results, I think it's safe to make this a minor release, not a major one. Thanks!

dlongley (approved):
Thanks!

lib/index.js
```diff
@@ -541,9 +552,8 @@ function _connect(options, callback) {
...
     logger.info('connecting to database: ' + _sanitizeUrl(options.url));
   }
-
-  let connectOptions = {...config.connectOptions};
-
+  const {writeConcern} = config.writeOptions;
+  let connectOptions = {...config.connectOptions, writeConcern};
   // convert legacy connect options
   if('socketOptions' in config.connectOptions) {
     connectOptions = {
```
davidlehn, line 559: [Late to the party] Is it ok that this will overwrite the options set above?

lib/index.js
```diff
@@ -58,6 +58,8 @@ api.db = null;
 // shared collections cache
 api.collections = {};
 
+// Note: api.writeOptions is deprecated
+// and will be removed in release 9.0
```
davidlehn, line 62: This sort of thing should perhaps be noted in the changelog, maybe detected and logged, and probably needs an issue so someone remember to remove it.

lib/index.js
aljones15: yeah @dlongley wants `writeConcern` to overwrite in connectOptions, but not in `gridFsBucket`.

lib/index.js
aljones15: PR for that here and good catch: https://github.com/digitalbazaar/bedrock-mongodb/pull/66

PR #66, opened 2021-02-11
Add deprecation notice to CHANGELOG.

aljones15:
see above


dlongley (approved):

davidlehn:
I suppose this is fine here.  In projects that have wider use it's good to make it more obvious what action needs to be taken.  As written you'd have to go look to see what the notice was.  In some other projects I've done something like:

```
## Notes
- `api.writeOptions` is deprecated and will be removed in a future release.  Switch to using [...].
```

Issue #47, opened 2020-06-21
Update from async package async/await.


aljones15:
This is a major breaking change, but this library is core to functionality and lags behind other libraries significantly. We have previously made changes to bedrock events to make then use `async` and `await`, but this library still uses an older callback based library.

Here some notes on the work needed (**NOTE: these notes were based off of version `6.0.0`**):

## `lib/index.js`

- [ ] [switch init to async](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L90)
- [ ] [await initDatabase in init](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L134-L142)
- [ ] [await openLocal](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L143-L153)
- [ ] [await dropCollections](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L154-L160)
- [ ] [await openCollections](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L161-L162)
- [ ] [await createIndex](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L163-L168)
- [ ] [await setupLocal](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L169-L176)
- [ ] [openCollections change to async and dependent libs will need to remove the callback](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L196)
- [ ] [openCollection remove async.each in favor of a for loop on unopened await createCollection](https://github.com/digitalbazaar/bedrock-mongodb/blob/83750f29103b8d77667ffc3fcebd1b13a893513d/lib/index.js#L205-L218)
- [ ] [mongo driver 3.5 createCollection docs (returns a promise if no callback is passed](http://mongodb.github.io/node-mongodb-native/3.5/api/Db.html#createCollection)
[...]

PR #71, opened 2021-07-11
Remove distributed ID collection and async clean up.

dlongley:
This PR should not introduce any API changes. It only removes an unused collection (that should have been removed long ago) and cleans up internal helper functions (e.g., they are now `async`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # bedrock-mongodb ChangeLog
 
+## 8.2.1 - 2021-xx-xx
+
+### Changed
+- Removed unused `distributedId` collection. This collection has not been
+  used since `idGenerator` was removed in 7.x (which was the only feature
+  that used it and it should have been removed at that time).
+
 ## 8.2.0 - 2020-02-22
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,14 @@
 # bedrock-mongodb ChangeLog
 
-## 8.2.1 - 2021-xx-xx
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -699,34 +660,37 @@ function _usesRoles(serverInfo) {
...
 
-function _dropCollections(callback) {
+async function _dropCollections() {
   if(bedrock.config.mongodb.dropCollections.collections === undefined) {
-    return callback(
-      new BedrockError(
-        'If bedrock.config.mongodb.dropCollection.onInit is specified, ' +
-        'bedrock.config.mongodb.dropCollection.collections must also ' +
-        'be specified.', 'InvalidConfiguration'));
+    new BedrockError(
```
davidlehn, line 665:
```suggestion
    return new BedrockError(
```

dlongley: This should be thrown, actually, fixed.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # bedrock-mongodb ChangeLog
 
+## 8.2.1 - 2021-xx-xx
```
davidlehn, line 3: Considering the extensive internal refactoring, perhaps 8.3.0 would be better. Leave some wiggle room in case something goes wrong. Maybe should also add a line item for the 'async' lib to async/await switch.

dlongley: Ok, updated.

davidlehn (approved):
Great cleanup, thanks!
I'm guessing this mostly or fully covers items here:
https://github.com/digitalbazaar/bedrock-mongodb/issues/47

PR #74, opened 2021-09-07
Use MongoClient result from _loginUser when adding user.

aljones15:
When adding a new user the new `async _loginUser` method now returns a fuil mongoDB client where previously it returned an object containing that client.

NOTE: this is a draft as I'm still testing the addUser functionality right now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # bedrock-mongodb ChangeLog
 
+## 8.4.1 - TBD
+
+### Changed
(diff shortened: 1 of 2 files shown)


dlongley (approved):
This seems like a correct fix to me.

dlongley:
It would be nice if we had an easy way to test this -- right now it requires manual testing, IIRC.

aljones15:
@dlongley I made an issue for testing with auth here: https://github.com/digitalbazaar/bedrock-mongodb/issues/75

Issue #84, opened 2022-07-08
Add call to get server info after logging in user (authenticated connection)


dlongley:
We need to ensure that the connection to the database works after trying to establish an authenticated connection. We may be able to ask for the server info again (or a similar call) after doing `loginUser`. Just returning after `loginUser` doesn't catch connection errors which leads to a lot of trouble.

davidlehn:
Maybe the minimal ping command is enough?

dlongley:
> Maybe the minimal ping command is enough?

Maybe. We'll need to experimentally confirm whatever solution we try.

mattcollier:
Here's something else that might be useful: https://mongodb.github.io/node-mongodb-native/3.7/api/Db.html#event:fullsetup

aljones15:
ok so some issues on issues:

Originally the mongodb driver has a method `db.authenticate`:

https://github.com/digitalbazaar/bedrock-mongodb/blob/10924b42c578857fefc998e872409ba5392e13bf/lib/index.js#L692-L696

This was ported over to `db.connect` as `db.authenticate` was deprecated when mongodb's node driver upgraded.

https://github.com/digitalbazaar/bedrock-mongodb/blob/9552919fd068b3423a8afc50b9ae33729105faea/lib/index.js#L672-L682

https://github.com/digitalbazaar/bedrock-mongodb/blob/9552919fd068b3423a8afc50b9ae33729105faea/lib/index.js#L722-L739

So I believe we can actually remove the function `loginUser` now as changes to the `_connect` function were made that use `auth` if a `username` and `password` are specified:

https://github.com/digitalbazaar/bedrock-mongodb/blob/d44a19830cac1d9634058a997ae790fa25929698/lib/authn.js#L138-L149

So basically, we have legacy code ported from an older driver that is resulting in multiple logins. That might be one of the underlying issues. This also means we need to check if auth is required before calling on `_connect`.

aljones15:
ok so further issue with this issue:

calls on `admin.serverInfo()` don't require authentication, but `listDatabases` does (which is weird because in mongo shell listDatabases will just return nothing if you're not authenticated. So serverInfo should come first, check the semver to make sure it matches the expected `semver` and then we need to check if auth is required after that.

mattcollier:
[...]

Issue #75, opened 2021-09-08
Add tests for Authentication


aljones15:
We have no tests for authentication and it's causing problems.

There is an existing issue suggesting using a docker container for production settings:

https://github.com/digitalbazaar/bedrock-mongodb/issues/39

The major challenges to testing auth is that existing tests test against mongodb setups on developer's computers. There are existing temp mongodb frameworks that might be able to solve this. Another option would be to start a docker mongodb container on a port different from the usual development port and give it an pre-set `mongod.conf` file with auth turned on. Then test against that container with the auth tests.

Auth tests also need to include adding a new user which does pause the CLI while you enter the admin details.
There might be more tests we need that involve auth.

dlongley:
Addressed by #85.

PR #85, opened 2022-07-15
Issue 84 get server info login

aljones15:
Addresses https://github.com/digitalbazaar/bedrock-mongodb/issues/84
https://github.com/digitalbazaar/bedrock-mongodb/issues/75

ToDo

- [x] Figure out what the socket options stuff does or did
- [x] Add tests that use a docker container with Auth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,10 @@
 # bedrock-mongodb ChangeLog
 
-## 10.1.1 -
+## 10.0.1 -
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
-- No longer pass in other undefined connectOptions.
-
-### Added
-- Added a new config option `checkServerDetails` that allows you to skip auth, server
-  version, & role checks.
+- No longer pass in other `ssl: undefined` in connectOptions.
 
 ## 10.0.1 -
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # bedrock-mongodb ChangeLog
 
-## 10.1.0 -
+## 10.1.1 -
+
+### Fixed
+- No longer pass in `authSource: undefined` as this causes connection strings to fail.
+- No longer pass in `ssl: undefined` as Mongo defaults to `ssl: true`.
 
 ### Added
 - Added a new config option `skipChecks` that allows you to skip auth, server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 10.1.1 -
 
 ### Fixed
-- No longer pass in `authSource: undefined` as this causes connection strings to fail.
-- No longer pass in `ssl: undefined` as Mongo defaults to `ssl: true`.
+- No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
+- No longer pass in other undefined connectOptions.
 
 ### Added
 - Added a new config option `skipChecks` that allows you to skip auth, server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - No longer pass in other undefined connectOptions.
 
 ### Added
-- Added a new config option `skipChecks` that allows you to skip auth, server
+- Added a new config option `checkServerDetails` that allows you to skip auth, server
   version, & role checks.
 
 ## 10.0.1 -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,22 +1,19 @@
 # bedrock-mongodb ChangeLog
 
-## 10.0.2 -
+## 10.0.1 -
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
 - No longer pass in `ssl: undefined` in connectOptions.
-- Use `db.databaseName` when logging connection success over `config.name`.
-
-## 10.0.1 -
-
-### Fixed
 - `openDatabase` now queries for serverInfo first.
 - `openDatabase` now correctly checks if auth is needed.
 - `openDatabase` no longer double logins in authenticated users.
+- `socketOptions` are being set correctly in connectOptions again.
+- Use `db.databaseName` when logging connection success over `config.name`.
 
 ### Added
-- `openDatabase` pings server on successful connect.
 - Github Actions now test on authenticated & unauthenticated databases.
+- `openDatabase` pings server on successful connect.
 
 ## 10.0.0 - 2022-04-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
 - No longer pass in `ssl: undefined` in connectOptions.
-- Use `db.databaseName` when logging connection success over `config.name`. 
-- Allow Mongo Driver to infer the database being connected to over using `config.name`.
+- Use `db.databaseName` when logging connection success over `config.name`.
 
 ## 10.0.1 -
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # bedrock-mongodb ChangeLog
 
-## 10.0.1 -
+## 10.0.2 -
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # bedrock-mongodb ChangeLog
 
+## 10.1.0 -
+
+### Added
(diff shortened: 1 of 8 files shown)


lib/authn.js
```diff
@@ -67,86 +37,63 @@ export async function openDatabase(options) {
...
-    ...config.authentication,
-    ...config.connectOptions,
-    ...config.writeOptions,
-  };
-
   // if the user specified a connection URL use it
-  let url = config.url;
-  if(!url) {
-    url = urls.create(config);
+  if(!opts.url) {
```
dlongley, line 59: This doesn't seem right ... should it be?:

```suggestion
  if(config.url) {
```

Or maybe:

```suggestion
  if(!opts.url && config.url) {
```

Otherwise it seems like it deviates from previous behavior and / or may cause breakage if `config.url` is not present? Or is there some other kind of correction happening?

lib/authn.js
```diff
@@ -67,86 +37,63 @@ export async function openDatabase(options) {
...
 
-  // authenticate against shared db
-  await loginUser({url, opts, auth});
-
-  logger.debug(
-    'database authentication succeeded: db=' + config.name +
-    ' username=' + config.username);
-
-  return {client, db};
+  //connect to database and get server info
```
dlongley, line 63:
```suggestion
  // connect to database and get server info
```

lib/authn.js
```diff
@@ -67,86 +37,63 @@ export async function openDatabase(options) {
...
-  // try to authenticate...
-
-  // if there is a username & password set, create an auth object for Mongo;
+  // if authRequired create an auth object for Mongo;
   // otherwise `auth` will be passed as `null` and success will rely on other
-  // config options
-  let auth = null;
-  if(config.username && config.password) {
-    auth = {
+  // config options such as the url for the server.
```
dlongley, line 46:
```suggestion
  // config options such as the url for the server
```

lib/authn.js
```diff
@@ -7,58 +7,28 @@ import {logger} from './logger.js';
...
-  }
-
-  if(!authRequired) {
-    // return early
-    logger.debug('database authentication not required');
-    return {client, db};
-  }
+  // connects to the mongo server unauthenticated
+  const {admin} = await _getUnauthenticatedDb({config: klona(config)});
+  // gets the serverInfo
```
dlongley, line 29:
```suggestion
  // do unauthenticated connection to mongo server to check
  // server compatibility and authn requirements
  const {admin} = await _getUnauthenticatedDb({config: klona(config)});
```

lib/authn.js
```diff
@@ -67,86 +37,63 @@ export async function openDatabase(options) {
       `MongoDB server version "${stringVersion}" is unsupported.`,
       'NotSupportedError');
   }
+  // makes an unauthenticated call on the server
```
dlongley, line 40:
```suggestion
  // makes an unauthenticated call to the server
```

lib/authn.js
```diff
@@ -156,3 +103,81 @@ function _usesRoles(serverInfo) {
     (serverInfo.versionArray[0] == 2 && serverInfo.versionArray[1] >= 6) ||
     (serverInfo.versionArray[0] > 2));
 }
+
+/**
+ * Determines if auth is required.
```
dlongley, line 108:
```suggestion
 * Determines if authn is required.
```

lib/authn.js
```diff
@@ -156,3 +103,81 @@ function _usesRoles(serverInfo) {
...
+ * https://mongodb.github.io/node-mongodb-native/3.7/api/Admin.html
+ * #listDatabases
+ * @private
+ * @param {object} options - Options to use.
+ * @param {object} options.config - The mongodb config.
+ * @param {object} options.admin - A Mongo driver Admin class.
+ *
+ * @returns {Promise<boolean>} Is authRequired?
+ */
+async function _authRequired({config, admin}) {
```
dlongley, line 120:
```suggestion
async function _isAuthnRequired({config, admin}) {
```

lib/authn.js
```diff
@@ -67,86 +37,63 @@ export async function openDatabase(options) {
       `MongoDB server version "${stringVersion}" is unsupported.`,
       'NotSupportedError');
   }
+  // makes an unauthenticated call on the server
+  // to see if auth is required
+  const authRequired = await _authRequired({config, admin});
```
dlongley, line 42:
```suggestion
  const authRequired = await _isAuthRequired({config, admin});
```

lib/authn.js
```diff
@@ -156,3 +103,81 @@ function _usesRoles(serverInfo) {
...
+      await admin.listDatabases();
+    } catch(e) {
+      if(e && e.code === MDBE_AUTHZ_FAILED) {
+        return true;
+      } else {
+        // some other error, abort
+        throw e;
+      }
+    }
+  }
```
dlongley, line 135:
```suggestion
  }
  try {
    // if listing databases fails on authz, then authentication is required
    await admin.listDatabases();
  } catch(e) {
    if(e?.code === MDBE_AUTHZ_FAILED) {
      return true;
    }
    // some other error, abort
    throw e;
  }
```

lib/authn.js
```diff
@@ -156,3 +103,81 @@ function _usesRoles(serverInfo) {
...
+    if(!semver.satisfies(version, config.requirements.serverVersion)) {
+      throw new BedrockError(
+        'Unsupported database version.',
+        'DatabaseError', {
+          url: urls.sanitize(config.url),
+          version,
+          required: config.requirements.serverVersion
+        });
+    }
+  }
```
dlongley, line 182:
```suggestion
  if(config.requirements.serverVersion &&
    !semver.satisfies(version, config.requirements.serverVersion)) {
    throw new BedrockError(
      'Unsupported database version.',
      'DatabaseError', {
        url: urls.sanitize(config.url),
        version,
        required: config.requirements.serverVersion
      });
  }
```

test/test.config.js
```diff
@@ -16,15 +16,15 @@ if(process.env.MONGODB_USERNAME && process.env.MONGODB_PASSWORD) {
   config.mongodb.username = process.env.MONGODB_USERNAME;
   config.mongodb.password = process.env.MONGODB_PASSWORD;
   const {connectOptions} = config.mongodb;
-  // this can be false
-  connectOptions.ssl = true;
+  // process.env.MONGODB_SSL should be 1 to test with SSL
+  connectOptions.ssl = Boolean(process.env.MONGODB_SSL) || false;
```
dlongley, line 20: As long as we're ensuring the tests run against SSL-enabled servers, this is fine.

dlongley (changes requested):
Thanks! A few stylistic nits and at least one question in the comments.

lib/authn.js
aljones15: ok so `openDatabase` copies the connection config stuff from config and passes those options to `_connect` now. There's a lot of clean up here.

test/test.config.js
aljones15: that is another issue actually (which we will need to address with changes to the docker images used to run mongodb servers). You have to specify a command when starting the mongoDB server with a valid SSL cert in order to run mongodb in SSL mode. I'm not sure when that changed from a minor error to a fatal error, but as of `mongodb 4.4` it no longer allows you to login if the server doesn't have a valid SSL certificate and SSL is required. This is fixable quite easily by passing a let's cert certificate to the mongodb docker image and then add the ssl cli command with a path to the certificate on start up. We don't do that currently which means projects that use docker mongodb images are probably not running with SSL. We also might want to add one of those change this for production guards to ensure in production we're running with SSL enabled. Basically good idea, but probably better address in a separate PR/ issue.

lib/authn.js
dlongley: Yes, the comment suggestion isn't covering only a single line, but the following block of code.

dlongley (approved):
Thanks! Please just apply that one other comment suggestion that elaborates on what the following code block is doing and we're good AFAICT from just reading the code. We'll need to experimentally confirm that these changes work with existing projects and solve the problem we're trying to solve. I imagine that even if they don't solve the existing problem, they are good still good changes to have since there's a lot of simplification.

mattcollier:
The approach being used here seems to be incompatible with MongoDB Atlas.  Atlas does not allow unauthenticated connections to the server in order to do the various checks that are happening here.  See my comment here: https://github.com/digitalbazaar/bedrock-mongodb/issues/84#issuecomment-1189535252

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+
+### Fixed
+- `openDatabase` now queries for serverInfo first.
+- `openDatabase` now correctly checks if auth is needed.
+- `openDatabase` no longer double logins in authenticated users.
+
+### Added
+- `openDatabase` pings server on successful connect.
+- Github Actions now test on authenticated & unauthenticated databases.
+
```
dlongley, line 20: This needs alphabetizing and merging. Are all these things still true now that the two branches have merged?

dlongley (changes requested):

dlongley (approved):

PR #87, opened 2022-07-19
Do not await client.close do log error.

aljones15:
https://github.com/digitalbazaar/bedrock-mongodb/issues/84

It looks like the code in the `finally` block executes before the `throw` in `catch` so we still try to close the connection, but we want to log the error still. Also of note:

```js
      await client.close(force).catch(error => {
        logger.error(
          'failed to close client used to initialize database', {error});
      });
```

not sure what happens you await a promise with a catch attached.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -189,7 +189,7 @@ async function _initDatabase() {
     // credentials)
     if(client) {
       const force = true;
-      await client.close(force).catch(error => {
+      client.close(force).catch(error => {
         logger.error(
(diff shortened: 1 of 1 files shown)


dlongley (commented):
Seems fine to not await the close here ... but my understanding is that other PRs may be replacing this. If we're not replacing this, my only concern is with how the underlying driver may behave if we haven't finished closing the client connection and we open a new one with different credentials. No need to discuss this further if we're simplifying the whole process and eliminating this.

PR #88, opened 2022-07-20
Remove authentication check

aljones15:
Removes the authentication check
Removes the serverVersion check
Removes the uses roles check
No unauthenticated connections to the server are made before trying to connect.

_This should fix issues with mongodb atlas which does not allow unauthenticated connections at all._

### DIFF
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -5,11 +5,8 @@ import * as bedrock from '@bedrock/core';
 import * as urls from './urls.js';
 import {logger} from './logger.js';
 import mongo from 'mongodb';
-import semver from 'semver';
-import {klona} from 'klona';
 
 const {MongoClient} = mongo;
-const {util: {BedrockError}} = bedrock;
 
 export async function openDatabase(options) {
   const config = bedrock.config.mongodb;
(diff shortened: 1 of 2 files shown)


davidlehn:
I don't know if this was aimed at a release.  The server target isn't always mongodb atlas.  Should these be optional rather than just removed in case they are needed in other cases?  Seems the version check at least can be important no matter what the target is.  It also has related code in `lib/config.js`.  And this would all need a breaking changelog entry.

aljones15:
> I don't know if this was aimed at a release. The server target isn't always mongodb atlas. Should these be optional rather than just removed in case they are needed in other cases? Seems the version check at least can be important no matter what the target is. It also has related code in `lib/config.js`. And this would all need a breaking changelog entry.

it sounds like the better idea is to add the `serverVersion` and `use roles` checks after login and then maybe do the unauthenticated check first.

dlongley:
If there is consensus that this is a breaking change (looks like it to me), we should look into doing the v4 driver upgrade at the same time. Doing a mongodb breaking change touches a lot of things and it would be better to do it once. Hopefully there are not many breaking changes w/v4.

aljones15:
> If there is consensus that this is a breaking change (looks like it to me), we should look into doing the v4 driver upgrade at the same time. Doing a mongodb breaking change touches a lot of things and it would be better to do it once. Hopefully there are not many breaking changes w/v4 (that affect us).

this could be done via a minor release with a new feature that allows you to skip the auth checks.
Something like:

```js
config.mongodb.skipAuthenicationCheck = true;
```

aljones15:
Closing in favor of a new feature that let's you avoid auth checks: https://github.com/digitalbazaar/bedrock-mongodb/pull/89

PR #89, opened 2022-08-22
Make checks optional

aljones15:
Adds a new feature `checkServerDetails` allowing mongodb atlas connections to occur with out authn, role, or server version checks.

I was able to connect to a mongodb atlas instance using this branch and just a connectString with the password in the connect string.

This is a minor release so that our implementations of the mongodb 3.0 driver works on Atlas.

Confirmation that using `db.databaseName` works:

```js
2022-08-29T15:24:57.159Z - debug: [bedrock-mongodb] database connection succeeded: db=bedrock_mongodb_test username=undefined workerPid=17100, workerId=
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,10 @@
 # bedrock-mongodb ChangeLog
 
-## 10.1.1 -
+## 10.0.1 -
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
-- No longer pass in other undefined connectOptions.
-
-### Added
-- Added a new config option `checkServerDetails` that allows you to skip auth, server
-  version, & role checks.
+- No longer pass in other `ssl: undefined` in connectOptions.
 
 ## 10.0.1 -
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # bedrock-mongodb ChangeLog
 
-## 10.1.0 -
+## 10.1.1 -
+
+### Fixed
+- No longer pass in `authSource: undefined` as this causes connection strings to fail.
+- No longer pass in `ssl: undefined` as Mongo defaults to `ssl: true`.
 
 ### Added
 - Added a new config option `skipChecks` that allows you to skip auth, server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 10.1.1 -
 
 ### Fixed
-- No longer pass in `authSource: undefined` as this causes connection strings to fail.
-- No longer pass in `ssl: undefined` as Mongo defaults to `ssl: true`.
+- No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
+- No longer pass in other undefined connectOptions.
 
 ### Added
 - Added a new config option `skipChecks` that allows you to skip auth, server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - No longer pass in other undefined connectOptions.
 
 ### Added
-- Added a new config option `skipChecks` that allows you to skip auth, server
+- Added a new config option `checkServerDetails` that allows you to skip auth, server
   version, & role checks.
 
 ## 10.0.1 -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
 - No longer pass in `ssl: undefined` in connectOptions.
-- Use `db.databaseName` when logging connection success over `config.name`. 
-- Allow Mongo Driver to infer the database being connected to over using `config.name`.
+- Use `db.databaseName` when logging connection success over `config.name`.
 
 ## 10.0.1 -
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # bedrock-mongodb ChangeLog
 
-## 10.0.1 -
+## 10.0.2 -
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # bedrock-mongodb ChangeLog
 
+## 10.1.0 -
+
+### Added
+- Added a new config option `skipChecks` that allows you to skip auth, server
+  version, & role checks.
+
 ## 10.0.1 -
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
-- No longer pass in other `ssl: undefined` in connectOptions.
+- No longer pass in `ssl: undefined` in connectOptions.
+- Use `db.databaseName` when logging connection success over `config.name`. 
+- Allow Mongo Driver to infer the database being connected to over using `config.name`.
 
 ## 10.0.1 -
 
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -24,7 +24,7 @@ export async function openDatabase(options) {
     ...options
   };
 
-  // if the user specified a connection URL use it
+  // if a `url` was not specified, create one from the `config`
   if(!opts.url) {
     opts.url = urls.create(config);
   }
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -88,7 +88,7 @@ async function _connect(options) {
     delete connectOptions.socketOptions;
   }
   const client = await MongoClient.connect(options.url, connectOptions);
-  const db = client.db(options.database);
+  const db = client.db();
   const ping = await db.admin().ping();
   logger.debug(
     'database connection succeeded: db=' + options.database +
@@ -167,7 +167,7 @@ async function _getUnauthenticatedDb({config}) {
   const url = `${urlParts.protocol}//${urlParts.host}${urlParts.pathname}`;
   const client = new MongoClient(url, opts);
   await client.connect();
-  const db = client.db(config.name);
+  const db = client.db();
   return {db, admin: db.admin()};
 }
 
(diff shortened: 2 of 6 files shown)


dlongley (changes requested):

lib/authn.js
```diff
@@ -23,23 +23,25 @@ export async function openDatabase(options) {
     writeOptions: {...config.writeOptions},
     ...options
   };
+  let authRequired = config.username && config.password;
+  if(!config.skipChecks) {
```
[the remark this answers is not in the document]
aljones15: It took a little work, but I did it here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/5aa582ec2b0368ef8b7e1dca9ddd6741916a0abf

lib/config.js
```diff
@@ -36,20 +36,21 @@ config.mongodb.authentication = {
   // used by MongoDB < 3.0
   // authMechanism: 'MONGODB-CR'
 };
-
+// if true role, server version, & authn checks are skipped
+// this is appears to be necessary on some mongo servers
+config.mongodb.skipChecks = false;
```
dlongley, line 41: `skipChecks` is not a good name because it isn't clear what it does. Can you suggest something more clear about its function?

dlongley (changes requested):
Before doing a full review here -- we should see if this solves the problem we're trying to solve.

lib/config.js
dlongley: You may be able to get a better name by inverting the flag to express what happens when the flag is *on* (and make that the new default) as well.

aljones15:
> Before doing a full review here -- we should see if this solves the problem we're trying to solve.

I have tried with a connection string at the free tier from mongodb atlas and it worked @mattcollier has the connection string that failed. So we need confirmation from him then?

lib/config.js
aljones15: I changed it to `checkServerDetails` here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/0db895e4351f66b66ad05b62ea3e4a0dd19db8fb

That seems better, the problem is we are skipping 3 checks: serverVersion, role check, and then the requiresAuthn check.

lib/authn.js
```diff
@@ -24,6 +24,11 @@ export async function openDatabase(options) {
     ...options
   };
 
+  // if the user specified a connection URL use it
```
dlongley, line 27: This is confusing ... because it's a NOT conditional. Does this mean?:

```suggestion
  // if a `url` was not specified, create one from the `config`
```

lib/authn.js
```diff
@@ -91,7 +81,7 @@ async function _connect(options) {
     delete connectOptions.socketOptions;
   }
   const client = await MongoClient.connect(options.url, connectOptions);
-  const db = client.db(options.database);
+  const db = client.db();
   const ping = await db.admin().ping();
   logger.debug(
     'database connection succeeded: db=' + options.database +
```
dlongley, line 87: Does `options.database` still always match now? i.e., can this log output be wrong / confusing?

lib/config.js
```diff
@@ -36,20 +36,18 @@ config.mongodb.authentication = {
...
   autoReconnect: false,
   useNewUrlParser: true,
-  // the db to authenticate against
-  authSource: undefined,
-  // we use the default for ssl
-  // set this to true if you want to require ssl
-  ssl: undefined,
   // promotes binary BSON values to native Node.js buffers
   promoteBuffers: true,
+  // FIXME: we need development and production servers to use SSL
```
dlongley, line 47: Why is this a FIXME?

lib/config.js
```diff
@@ -63,7 +61,7 @@ config.mongodb.writeOptions = {
 
 config.mongodb.requirements = {};
 // server version requirement with server-style string
-config.mongodb.requirements.serverVersion = '>=4.2';
+config.mongodb.requirements.serverVersion = '>=4.4';
```
dlongley, line 64: This is a breaking change, isn't it? Do we need to do this for this release?

dlongley (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 10.0.1 -
```
dlongley, line 3:
```suggestion
## 10.0.2 -
```

Right?

lib/config.js
aljones15: ok so in order to make SSL: true a default now it would be a breaking change (especially in development where the local mongodb server people use are usually not using SSL). I've talked to @mattcollier about this and basically we need to revise the docker images used for mongodb in development to take in a PEM ed25519 certificate for SSL.

lib/authn.js
aljones15: good catch this should use the db object for the name in all cases.

lib/config.js
aljones15: Ok reverted here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/65d4986a44b2985a5f3c22ba2a3a67833be5c6e1

lib/authn.js
aljones15: I corrected this here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/881884e81b6cacfa403d8ffcc0ebe2c0f29dd254 but there might be other places so a more thorough review of the lib is needed.

lib/config.js
aljones15: Just to make this more clear if `ssl: true` is an option, then mongodb servers with out ssl configured will hang and fail to connect. So turning it to true by default is definitely breaking.

lib/config.js
aljones15: This feature was removed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/7718dca684f277d75621c137c5c9580b77870f0c

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### Fixed
 - No longer pass in `authSource: undefined` as this causes connection strings with auth to fail.
-- No longer pass in other `ssl: undefined` in connectOptions.
+- No longer pass in `ssl: undefined` in connectOptions.
+- Use `db.databaseName` when logging connection success over `config.name`. 
+- Allow Mongo Driver to infer the database being connected to over using `config.name`.
```
aljones15, line 9: This last line is not a change actually as this is the behavior in the current release: https://github.com/digitalbazaar/bedrock-mongodb/blob/d44a19830cac1d9634058a997ae790fa25929698/lib/authn.js#L148-L149

lib/config.js
dlongley: That's fine, but this doesn't sound like a FIXME. It should just be a comment about other these options that can be set. The comment should remove the "FIXME" portion and then indicate that the options should be set in production and that they can't be set to undefined for `<enter reason here>`. If we need another comment that indicates a future breaking change may set them to some other default, that can go there too.

test/test.config.js
```diff
@@ -7,29 +7,39 @@ import path from 'node:path';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+const convertToBoolean = (envVariable = '') => {
+  if(/^(1|true)$/.test(envVariable)) {
+    return true;
+  }
+  return false;
+};
```
dlongley, line 15:
```suggestion
const convertToBoolean = (envVariable = '') => /^(1|true)$/.test(envVariable);
```

test/test.config.js
```diff
@@ -7,29 +7,39 @@ import path from 'node:path';
...
+// set the env variable to anything else to make them false
+if(process.env.MONGODB_SSL) {
+  connectOptions.ssl = convertToBoolean(process.env.MONGODB_SSL);
+}
+if(process.env.MONGODB_TLS) {
+  connectOptions.tls = convertToBoolean(process.env.MONGODB_TLS);
 }
+if(process.env.MONGODB_AUTHSOURCE) {
+  connectOptions.authSource = process.env.MONGODB_AUTHSOURCE;
+}
```
dlongley, line 33: Does all this "just work" for existing devs running stuff locally?

test/test.config.js
```diff
@@ -7,29 +7,39 @@ import path from 'node:path';
...
+
+// used for testing url only connections
+config.mongodb.url = process.env.MONGODB_URL;
+// this can safely be undefined
+connectOptions.replicaSet = process.env.MONGODB_REPLICASET;
+
+config.mongodb.username = process.env.MONGODB_USERNAME;
+config.mongodb.password = process.env.MONGODB_PASSWORD;
```
dlongley, line 41: Same question here ... does this just continue to work for devs running things locally? It seems like it would break.

dlongley (changes requested):

test/test.config.js
aljones15: This is test data that is only set for the tests and the local tests running without auth are fine. I did have to explicitly add `MONGODB_AUTHSOURCE=admin` so the github auth cli tests continue to work. That is more in line though with our production environment and also allows us to simulate having the wrong `authSource` quickly when debugging connection strings / options. Previously the `authSource` when testing defaulted to `admin` if `password` and `username` were provided: https://github.com/digitalbazaar/bedrock-mongodb/blob/issue-84-get-server-info-login/test/test.config.js

lib/config.js
aljones15: I changed it to recommend setting either `tls` or `ssl` to true in production: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/60ee8241d1355bfcc535ec93abaab4154422a4fe

test/test.config.js
dlongley: It just seems like these should only be set if the environment variable is also set. Otherwise someone modifying their local test config will be confused when it doesn't work. So please add `if` conditionals here -- we shouldn't override something someone has set in the config unless the environment variable is actually set.

dlongley (approved):
This is ready to go once the last concern is addressed. Let me know when it's ready and I'll do one more pass and merge and release. Thanks!

test/test.config.js
aljones15: Addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/3f4adaf8125c1a91910df261065578b738654944

test/test.config.js
aljones15: Good idea and addressed here: https://github.com/digitalbazaar/bedrock-mongodb/pull/89/commits/3f4adaf8125c1a91910df261065578b738654944

dlongley (approved):

PR #90, opened 2022-08-31
Patch unauthenticated check

aljones15:
This further cleans up code in `authn.js`, but is untested with major projects.

This is not a priority patch and can wait. It is more style and a minor bug fix.

### DIFF
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -13,20 +13,21 @@ const {MongoClient} = mongo;
 const {util: {BedrockError}} = bedrock;
 
 export async function openDatabase(options) {
-  const config = bedrock.config.mongodb;
+  // copy the config so we don't mutate something while connecting
+  const config = klona(bedrock.config.mongodb);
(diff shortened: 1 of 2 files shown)


dlongley:
Needs rebase.

lib/authn.js
```diff
@@ -143,21 +143,20 @@ function _addAuthOptions({options, config}) {
...
  * @private
  * @param {object} options - Options to use.
- * @param {object} options.config - Config options.
+ * @param {object} options.options - Connection options.
  *
  * @returns {Promise<object>} Returns an object with the db and admin.
  */
-async function _getUnauthenticatedDb({config}) {
+async function _getUnauthenticatedDb({options}) {
   // only use the config stuff with out auth credentials
```
dlongley, line 151: This comment is not correct any more, right? And is this the right thing to do since this is supposed to be an authenticated DB connection?

dlongley (changes requested):

PR #92, opened 2022-09-02
MAJOR: Upgrade to node mongodb driver 4

aljones15:
This is the initial PR for the mongo driver upgrade to 4. This is a major release. New Features and Subtractions will be made in PRS on this PR in order to make code review more thorough.

- [ ] Review code scanning for untested uses of `@bedrock/mongodb`
- [ ] Retest using mongo server 4, 5, & 6
- [ ] Replace branches in all PRs linked to this PR with releases of respective libraries.

**DRAFT PR NOT READY FOR RELEASE**

### DIFF
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -128,7 +128,7 @@ async function _isAuthnRequired({config, admin}) {
 
 function _addAuthOptions({options, config}) {
   options.connectOptions.auth = {
-    user: config.username,
+    username: config.username,
     password: config.password
   };
   // authSource is the database to authenticate against
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - **BREAKING**: Upgrade to `mongodb: ^4.9.0`
 
 ### Removed
-- Remove `config.adminPrompt`.
+- Remove unused `config.adminPrompt`.
 
 ## 10.1.1 -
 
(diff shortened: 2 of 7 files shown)


lib/authn.js
```diff
@@ -128,7 +128,7 @@ async function _isAuthnRequired({config, admin}) {
 
 function _addAuthOptions({options, config}) {
   options.connectOptions.auth = {
-    user: config.username,
+    username: config.username,
```
dlongley, line 131: Reminder: We will be removing all these options in the next major release and just using `url`.

aljones15: I actually don't recall agreeing to that, I recall suggesting that we follow the mongo upgrade advice and either use object auth or url auth, but never a mix of both. Regardless this commit fixes the auth tests which saves us 30 minutes of tests that weren't timing out.

dlongley: Having just one way to do things is better. Unless there's some extra feature we get (which we will need to debate the usefulness) by having two ways, we should just have one way. The simplest way to do that when configuring is with a single `url` value.

aljones15: As long as we have a `connectOptions` someone can use an object to authorize. Anyways, let's discuss this more in the next PR which removes:

1. All defaults for `authSource`
2. `username` and `password` are removed from config
3. All of the auth related object construction is gone

dlongley:
Closing in favor of #107.

PR #94, opened 2022-09-02
Mongo 4 no roles check

aljones15:
1. Removes the roles check for `./lib/authn.js`
2. Throws if the server version is less than 3.6.0 (this was tested and was the lowest mongo node 4 goes)
3. Ups the default server version to 4.4

TODO:

- [x] Test with mongodb server v3
- [x] Test with mongodb server v2

Node Driver compatibility matrix is here: https://www.mongodb.com/docs/drivers/node/current/compatibility/#mongodb-compatibility

3.6 seems to be the lowest.

### DIFF
--- a/lib/authn.js
+++ b/lib/authn.js
@@ -35,13 +35,6 @@ export async function openDatabase(options) {
   const serverInfo = await admin.serverInfo(null);
   _checkServerVersion({serverInfo, config});
 
-  // check if server supports roles; if not, can't authenticate
-  if(!_usesRoles(serverInfo)) {
-    const stringVersion = serverInfo.versionArray.join('.');
-    throw new BedrockError(
-      `MongoDB server version "${stringVersion}" is unsupported.`,
-      'NotSupportedError');
-  }
   // makes an unauthenticated call to the server
   // to see if auth is required
   const authRequired = await _isAuthnRequired({config, admin});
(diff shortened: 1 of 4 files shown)


lib/authn.js
```diff
@@ -170,11 +156,22 @@ function _checkServerVersion({serverInfo, config}) {
...
   });
+  // if the server is not at least 3.6.0 throw
+  // as mongo node driver 4 only supports 3.6.0 and higher
+  if(!semver.gte(version, '3.6.0')) {
+    throw new BedrockError(
+      'Unsupported database version.',
+      'NotSupportedError', {
+        url: urls.sanitize(config.url),
+        version,
+        required: '>= 3.6.0'
```
mattcollier, line 167: Isn't a different version being required by code changes elsewhere in this PR?

lib/config.js
```diff
@@ -56,12 +57,11 @@ config.mongodb.writeOptions = {
     w: 'majority',
     j: true,
   },
-  forceServerObjectId: true,
 };
 
 config.mongodb.requirements = {};
 // server version requirement with server-style string
-config.mongodb.requirements.serverVersion = '>=4.2';
+config.mongodb.requirements.serverVersion = '>=4.4';
```
mattcollier, line 64: Is 4.4 a requirement?  We do still have some clusters on Atlas that are 4.2.x.

lib/authn.js
aljones15: The driver only goes down to 3.6 now: https://www.mongodb.com/docs/drivers/node/current/compatibility/#mongodb-compatibility

So 3.6 is the hard limit on server version.

lib/config.js
aljones15: Ok so `4.4` is the recommended min version, you can set the serverVersion as low as 3.6. See the compatibility matrix here: https://www.mongodb.com/docs/drivers/node/current/compatibility/#mongodb-compatibility

dlongley:
Closing in favor of #107.

PR #95, opened 2022-10-28
Mongo driver 4 rc fix duplicate error

aljones15:
1. Fixes the Duplicate check
2. Adds a check for `MongoError`.

### DIFF
--- a/lib/exceptions.js
+++ b/lib/exceptions.js
@@ -1,73 +1,17 @@
 /*!
  * Copyright (c) 2012-2022 Digital Bazaar, Inc. All rights reserved.
  */
+
+import {MongoError} from 'mongodb';
+
 // Mongo DB Errors
 // Errors outside of the Mongo class of Errors
 export const WRITE_CONCERN_ERROR = 'WriteConcernError';
 export const WRITE_ERROR = 'WriteError';
-
-export const MDBE_API_ERROR = 'MongoAPIError';
-export const MDBE_AWS_ERROR = 'MongoAWSError';
-export const MDBE_BATCH_REEXECUTION_ERROR = 'MongoBatchReExecutionError';
-export const MDBE_BULK_WRITE_ERROR = 'MongoBulkWriteError';
-export const MDBE_CHANGE_STREAM_ERROR = 'MongoChangeStreamError';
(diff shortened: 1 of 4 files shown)


lib/exceptions.js
```diff
@@ -1,16 +1,80 @@
...
+  MDBE_SYSTEM_ERROR,
+  MDBE_TAILABLE_CURSOR_ERROR,
+  MDBE_TOPOLOGY_CLOSED_ERROR,
+  MDBE_TRANSACTION_ERROR,
+  MDBE_UNEXPECTED_SERVER_RESPONSE_ERROR,
+  MDBE_SERVER_ERROR,
+  MDBE_WRITE_CONCERN_ERROR,
+  WRITE_CONCERN_ERROR,
+  WRITE_ERROR
+]);
```
dlongley, line 77: -1 to adding all these, we should only add what we need (and we haven't needed any of these) -- otherwise it's just more to maintain and test.

dlongley (changes requested):

lib/exceptions.js
aljones15: That's a good point. There might be an automated way of adding these though. I'll look into it and if not cut them down to something more manageable. I also think not all of these errors are root errors, many of them are the cause of a `MongoError` I believe.

aljones15: ok so I trimmed this down to just the two errors that are kind of sort of thrown by the driver that are not `MongoError`s. Then I added a new function `assertMongoError` here: https://github.com/digitalbazaar/bedrock-mongodb/pull/95/commits/43ab9b0f1c3503dc31a8a8847e918c0c6638b70c
and here:
https://github.com/digitalbazaar/bedrock-mongodb/pull/95/commits/94bf5291c7994e72c0a7943895dbffaa6d86c69c

This method should work as the node mongo driver is now written using typescript so there are a lot of Abstract base classes and inheritance based type checking structures that ultimately result in 3 cases: either you get a `MongoError`, a `WriteError` or `WriteConcernError`. For the most part you only get `MongoError`, but documents can contain `WriteError` and `MongoError` can contain both `WriteError` or `WriteConcernError` so there might be edge cases were we're asserting on the cause of a `MongoError` or on the `writeErrors` property of `MongoError`.

PR #98, opened 2023-01-05
Remove deprecated option adminPrompt.

aljones15:
This removes `adminPrompt` which used to pause bedrock if the user for the database wasn't found.
It would then ask for the admin user and password and then add users to the database. `adminPrompt` was deprecated at some point in a previous major release, but the config option wasn't removed. So this PR is basically clean up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - **BREAKING**: Upgrade to `mongodb: ^4.9.0`
 
 ### Removed
-- Remove `config.adminPrompt`.
+- Remove unused `config.adminPrompt`.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 ### Changed
 - **BREAKING**: Upgrade to `mongodb: ^4.9.0`
 
+### Removed
+- Remove `config.adminPrompt`.
```
dlongley, line 9:
```suggestion
- Remove unused `config.adminPrompt`.
```

dlongley (approved):

PR #101, opened 2023-12-13
Upgrade to MongoDB Driver v6

mattcollier:
Thanks to @aljones15 for previous work in exploring the driver upgrade.

This implementation:
- Works with and without authentication in the local test suite.
- Works with existing test suites in relying libs without modification to the test framework/config.
- Works with MongoDB Atlas in deployments without any changes to connect strings or config.
- There should be no breaking changes in terms of the connection/authentication.
- There are breaking changes related to the use of various driver APIs
  - The `insert` APIs no longer return the document that was inserted.

This has been tested with one top level app.  I will be plugging it into additional Bedrock libs before release.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright (c) 2012-2022 Digital Bazaar, Inc. All rights reserved.
+ * Copyright (c) 2012-2024 Digital Bazaar, Inc. All rights reserved.
  */
 import * as bedrock from '@bedrock/core';
 import * as urls from './urls.js';
--- a/lib/index.js
+++ b/lib/index.js
@@ -81,12 +81,8 @@ export async function openCollections(names) {
   // open the collections
   logger.debug('opening collections', {collections: unopened});
   const collections = {};
-  const {writeConcern} = _db.options;
   await Promise.all(unopened.map(async name => {
-    // Note: We only pass `{writeConcern}` here to get around a bug in mongodb
-    // node driver 3.6.4 where `writeConcern` from `db` is not passed to
-    // collection
-    collections[name] = await _db.collection(name, {writeConcern});
+    collections[name] = await _db.collection(name);
   }));
 
   // merge results into collection cache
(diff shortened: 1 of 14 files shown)


test/mocha/10-api.js
```diff
@@ -406,6 +406,26 @@ describe('api', function() {
       }
       should.exist(result);
       should.not.exist(error);
+      result.should.have.keys(['acknowledged', 'insertedId']);
+    });
+    it('should properly promote binary values to buffers', async function() {
```
aljones15, line 411: this is cool thanks for adding this.

aljones15:
has this been tested for interoperability with the multiple backend libs that use `@bedrock/mongodb`? the big issue holding up the v4 driver was changes to how the modified results count were returned.

mattcollier:
@aljones15 the types of changes you mentioned related to driver API return value changes etc will still need to be addressed in relying libs.

aljones15:
> @aljones15 the types of changes you mentioned related to driver API return value changes etc will still need to be addressed in relying libs.

ok well maybe those existing PRs can be revised to work with driver 6. thanks for this PR.

lib/index.js
```diff
@@ -81,12 +81,8 @@ export async function openCollections(names) {
   // open the collections
   logger.debug('opening collections', {collections: unopened});
   const collections = {};
-  const {writeConcern} = _db.options;
   await Promise.all(unopened.map(async name => {
-    // Note: We only pass `{writeConcern}` here to get around a bug in mongodb
-    // node driver 3.6.4 where `writeConcern` from `db` is not passed to
-    // collection
```
dlongley, line 88: Presumably this bug has been fixed now.

dlongley (approved):
Thanks!

dlongley:
Some of the changes in this PR for v6 need to be adjusted (`writeConcern`) per #106. We might also just want a fresh PR since there are many conflicts with this one now.

dlongley (changes requested):
See #106 and latest comment here.

dlongley:
Closing in favor of #107, which includes some of the non-conflicting commits from this PR.

PR #107, opened 2025-03-05
Update to driver v6.

dlongley:
Replaces all other PRs related to driver upgrades.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,3 +1,20 @@
+/*!
+ * Copyright 2012 - 2025 Digital Bazaar, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
 module.exports = {
   root: true,
   env: {
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,30 +1,15 @@
-/*!
- * Copyright 2012 - 2024 Digital Bazaar, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


lib/config.js
```diff
@@ -79,7 +75,7 @@ config.mongodb.writeOptions = {
 
 config.mongodb.requirements = {};
 // server version requirement with server-style string
-config.mongodb.requirements.serverVersion = '>=4.2';
+config.mongodb.requirements.serverVersion = '>=5';
```
davidlehn, line 78: The docs and tests are using v6.  Is there a reason to have this lower?

.eslintrc.cjs
```diff
@@ -15,16 +15,18 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-
```
davidlehn, line 18: @BigBlueHat I recall the license tooling added these blank lines?  Do they need to be preserved?

package.json
```diff
@@ -37,12 +36,13 @@
...
   "devDependencies": {
-    "eslint": "^7.13.0",
-    "eslint-config-digitalbazaar": "^2.8.0",
-    "eslint-plugin-jsdoc": "^37.9.7",
-    "jsdoc-to-markdown": "^7.1.1"
+    "eslint": "^8.57.1",
+    "eslint-config-digitalbazaar": "^5.2.0",
+    "eslint-plugin-jsdoc": "^50.6.3",
+    "eslint-plugin-unicorn": "^55.0.0",
+    "jsdoc-to-markdown": "^9.1.1"
```
davidlehn, line 43: remove unused dependency
```suggestion
    "eslint-plugin-unicorn": "^55.0.0"
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # bedrock-mongodb ChangeLog
 
+## 11.0.0 - 2025-mm-dd
+
+### Changed
+- **BREAKING**: Use mongodb driver 6.x.
+- **BREAKING**: Update error names to match bedrock best practice.
```
davidlehn, line 7: Does this need to be more verbose with a list of old to new mappings?  I'm not sure how pervasive, if at all, checking for specific names might be in calling code.  Similarly may need to mention some error messages/details may have changed?

lib/config.js
dlongley: Yes, there is software out there still using 5 for which this works -- we can bump it further in the future to avoid additional blockers on getting this update through.

.eslintrc.cjs
dlongley: No, in fact, the licensing tool should not be inserting them.

davidlehn: I don't think it was a choice, similar to the extra spaces in the copyright date range.  I was wondering if some part of that tooling would be grumpy if blank lines are gone.

